Issue #24, opened 2017-04-01
Remove redundant DummyToken contracts + update TokenRegistry migration.


willwarren89:
Rather than deploying DummyTokenA, DummyTokenB, DummyEtherToken and DummyProtocolToken as four separate contracts, instead deploy four identical "DummyToken" contracts and register each one in the TokenRegistry. For reference, [see here](https://github.com/0xProject/0x/blob/8a82d6f54970d84041e0ed984c2650aca5d6ea0a/src/client/truffle/migrations/3_register_tokens.js).

Customize contract deployment according to the network being deployed to, as in [this example](https://github.com/0xProject/0x/blob/8a82d6f54970d84041e0ed984c2650aca5d6ea0a/src/client/truffle/migrations/3_register_tokens.js). Deploy an actual EtherToken rather than a dummy version, for Ropsten/Kovan and live networks (or use Maker's EtherToken for the live network).

PR #34, opened 2017-04-22
Token deployment

abandeali1:
This PR:
-Consolidates all DummyTokens into a single contract with a more generic interface, and then deploys multiple instances of DummyToken.
-Adds a file `migrations/tokens.js` to configure tokens to deploy and add to TokenRegistry on different networks
-Reorganizes deployment steps
-Loads all DummyToken instances in tests from the TokenRegistry

Resolves issue #24


migrations/2_deploy_independent_contracts.js
```diff
@@ -19,25 +18,19 @@ module.exports = (deployer, network, accounts) => {
...
     deployer.deploy([
       [MultiSigWallet, config.owners, config.confirmationsRequired, config.secondsRequired],
       Proxy,
-      [DummyProtocolToken, 0],
       TokenRegistry,
-    ])
-    .then(() => deployer.deploy(Exchange, DummyProtocolToken.address, Proxy.address))
-    .then(() => Proxy.deployed())
-    .then(proxy => proxy.addAuthorizedAddress(Exchange.address, { from: accounts[0] }));
+      DummyEtherToken,
```
fabioberger: Question: That Ether token will we be using on the livenet? An existing one that is already deployed or our own?

migrations/tokens.js
```diff
@@ -0,0 +1,108 @@
+module.exports = {
```
fabioberger: Can we put this file and `multisig_config_sample.js` into a `configs` folder nested under the migrations folder? I think it would be cleaner to have them in a separate file from the actual migration files.

migrations/3_register_tokens.js
```diff
@@ -0,0 +1,76 @@
...
+        const totalSupply = 100000000;
+        return Promise.all(tokenSymbols.map(sym => DummyToken.new(
+          tokens[sym].name,
+          tokens[sym].symbol,
+          tokens[sym].decimals,
+          totalSupply
+        ))).then(instances => {
+          const weth = {
+            address: DummyEtherToken.address,
+            name: 'ETH Wrapper Token',
```
fabioberger: Can we name this `Ether Token` instead? We can add an explainer to the demo app elucidating the 1-to-1 convertibility between the two.

migrations/3_register_tokens.js
```diff
@@ -0,0 +1,76 @@
...
+            symbol: 'WETH',
+            url: '',
+            decimals: 18,
+            ipfsHash: '0x0',
+            swarmHash: '0x0',
+          };
+          return Promise.all(instances.map((tokenContract, i) => tokenRegistry.addToken(
+            tokenContract.address,
+            tokens[tokenSymbols[i]].name,
```
fabioberger: Cleaner would be to declare token once at the top of the loop body and reference it multiple times:

```
return Promise.all(instances.map((tokenContract, i) => {
  const token = tokens[tokenSymbols[i]];
  return tokenRegistry.addToken(
    tokenContract.address,
    token.name,
    token.symbol,
    ...
  )
```

migrations/3_register_tokens.js
```diff
@@ -0,0 +1,76 @@
...
+        address: ProtocolToken.address,
+        name: '0x Protocol Token',
+        symbol: 'ZRX',
+        url: 'https://www.0xproject.com/',
+        decimals: 18,
+        ipfsHash: '0x0',
+        swarmHash: '0x0',
+      };
+      return Promise.all(tokenSymbols.map(sym => tokenRegistry.addToken(
+        tokens[sym].address,
```
fabioberger: Same as above comment for here

migrations/3_register_tokens.js
```diff
@@ -0,0 +1,76 @@
...
+            weth.name,
+            weth.symbol,
+            weth.url,
+            weth.decimals,
+            weth.ipfsHash,
+            weth.swarmHash
+          )));
+        });
+      }
+      const zrx = {
```
fabioberger: Let's put this in a `else` block. It was less clear what was happening in `live` and `!live` environments from quickly glancing at the file. Also, this clarity will avoid someone making mistakes in the future.

migrations/2_deploy_independent_contracts.js
abandeali1: We'll be using the existing Maker implementation (I believe Gnosis is using the same one as well).

migrations/3_register_tokens.js
```diff
@@ -10,7 +10,7 @@ module.exports = (deployer, network) => {
   deployer.then(() =>
     TokenRegistry.deployed().then(tokenRegistry => {
       if (network !== 'live') {
-        const totalSupply = 100000000 * (10 ** 18);
+        const totalSupply = 100000000 * Math.pow(10, 18); // eslint-disable-line no-restricted-properties
```
willwarren89: Solidity :D

contracts/tokens/DummyToken.sol
```diff
@@ -0,0 +1,33 @@
...
+    symbol = _symbol;
+    decimals = _decimals;
+    totalSupply = _totalSupply;
+    balances[msg.sender] = _totalSupply;
+  }
+
+  function setBalance(address _target, uint _value) onlyOwner {
+    uint currBalance = balanceOf(_target);
+    if (_value < currBalance) {
+      totalSupply -= (currBalance - _value);
```
fabioberger: Let's use `safeSub` here

contracts/tokens/DummyToken.sol
```diff
@@ -0,0 +1,33 @@
...
+    } else {
+      totalSupply += (_value - currBalance);
```
fabioberger: Let's use `safeAdd` and `safeSub` here

migrations/config/tokens.js
```diff
@@ -0,0 +1,108 @@
...
+      name: 'Digix DAO Token',
+      symbol: 'DGD',
+      url: 'https://www.dgx.io/',
+      decimals: 18,
+      ipfsHash: '0x0',
+      swarmHash: '0x0',
+    },
+    GNT: {
+      address: '0xa74476443119a942de498590fe1f2454d7d4ac0d',
+      name: 'Golem Netwok Token',
```
fabioberger: Typo: Netwok -> Network

test/js/exchange/core.js
```diff
@@ -3,70 +3,77 @@ require('source-map-support/register');
...
 const Exchange = artifacts.require('./Exchange.sol');
 const Proxy = artifacts.require('./Proxy.sol');
-const DummyTokenA = artifacts.require('./tokens/DummyTokenA.sol');
-const DummyTokenB = artifacts.require('./tokens/DummyTokenB.sol');
-const DummyProtocolToken = artifacts.require('./tokens/DummyProtocolToken.sol');
+const DummyToken = artifacts.require('./tokens/DummyToken.sol');
+const TokenRegistry = artifacts.require('./TokenRegistry.sol');
 
 const assert = require('assert');
-const expect = require('chai').expect;
```
fabioberger: Nice! Great job getting rid of this dependency 👍

PR #35, opened 2017-04-22
Test util refactor

fabioberger:
The PR:
- Refactors the test utils in the following ways:
  - Remove level of indirection introduced by util index.js
  - Standardizes util modules to either be an exported class or object
  - Renamed suffix from `util` to `wrapper` where appropriate
  - Refactor such that, as much as possible, logic lives close to the data it manipulates and accesses. This means objects with instance variables and methods.
- Switches over test suite to using [Lodash](https://lodash.com/docs/4.17.4) for all our utility method needs. Useful ones already in use are `_.map`, `_.each`, `_.isBoolean`, `_.assign`, `_.includes`, `_.find`. These methods increase code readability and consistency, with the added benefit of each of these utilities being well documented in [the docs](https://lodash.com/docs/4.17.4).
- Gets rid of almost all param objects being passed to methods. We should standardize around not using this approach to method passing and instead pass regular named params. Regular named params increase code readability and any benefits of param objects will be superseded by actual type definitions when we switch to TypeScript.
- Postfix async methods with the keyword `async` since forgetting to call `await` on these methods causes difficult to detect failures.
- Standardize filenames to use snakecase instead of camelcase because of the lower guarentee that capitalization of filenames is respected in all environments.

Additional future improvements planned for a subsequent PR:
- Port the test suite to typescript. This will help us detect a huge number of bugs and type inconsistencies in our code, and will also allow us to create defined types for more complex objects we are currently passing around with insufficient validation.
- Get rid of relative path hell `../../..` in module imports (side-benefit of switching to TypeScript!)


abandeali1:
Very solid refactor. I do think a few things could be tweaked to make the code more reusable without hurting test readability (e.g making all the `async` methods return promises instead).

util/order.js
```diff
@@ -0,0 +1,82 @@
...
+      const pubKey = ethUtil.ecrecover(msgHash, v, ethUtil.toBuffer(r), ethUtil.toBuffer(s));
+      const recoveredAddress = ethUtil.bufferToHex(ethUtil.pubToAddress(pubKey));
+      return recoveredAddress === this.params.maker;
+    } catch (err) {
+      return false;
+    }
+  }
+  async signAsync() {
+    const orderHash = this.getOrderHash();
+    const msgHash = ethUtil.hashPersonalMessage(orderHash);
```
abandeali1: We should note that this will cause order signatures to be invalid if created using geth or Parity. I'm fine with this until we add either a client config file or built in client-detection, though.

fabioberger: Why will it fail for geth and parity?

abandeali1: `eth_sign` automatically hashes the data to sign with the Ethereum signed message header in geth and (I believe) Parity. So if using geth or Parity, we would be hashing the header in twice.

fabioberger: I see, thanks for the heads up! Will add a comment for now and in a future refactor we'll add a client config file.

Issue #32, opened 2017-04-19
(Exchange.sol) Log error events when failing gracefully


willwarren89:
**Problem**

The Exchange contract can fail by throwing an exception, `throw`, or fail gracefully via `return 0` depending upon the error condition that is encountered. When Exchange fails gracefully it *also* fails silently which makes it difficult for a front end to know exactly what caused the error and provide feedback to the user.

**Proposed Solution**

Add `event LogError('Error description goes here.')` to Exchange. Fire the event with an appropriate error description whenever a graceful failure occurs.

willwarren89:
Including orderHash: `event LogError(string indexed error, bytes32 orderHash)`

Looks like this breaks a few of our tests.

PR #36, opened 2017-04-23
Log error

willwarren89:
See https://github.com/0xProject/contracts/issues/32

Tests are currently failing. I decided against attempting to fix the tests as they are being refactored in a separate branch.

### DIFF
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -26,12 +26,12 @@ import "./base/SafeMath.sol";
 contract Exchange is SafeMath {
 
   // Error Codes
-  uint8 constant ERROR_FILL_EXPIRED = 1;           // Order has already expired
-  uint8 constant ERROR_FILL_NO_VALUE = 2;          // Order has already been fully filled or cancelled
-  uint8 constant ERROR_FILL_TRUNCATION = 3;        // Rounding error too large
-  uint8 constant ERROR_FILL_BALANCE_ALLOWANCE = 4; // Insufficient balance or allowance for token transfer
-  uint8 constant ERROR_CANCEL_EXPIRED = 5;         // Order has already expired
-  uint8 constant ERROR_CANCEL_NO_VALUE = 6;        // Order has already been fully filled or cancelled
+  uint8 constant ERROR_FILL_EXPIRED = 0;           // Order has already expired
+  uint8 constant ERROR_FILL_NO_VALUE = 1;          // Order has already been fully filled or cancelled
+  uint8 constant ERROR_FILL_TRUNCATION = 2;        // Rounding error too large
+  uint8 constant ERROR_FILL_BALANCE_ALLOWANCE = 3; // Insufficient balance or allowance for token transfer
+  uint8 constant ERROR_CANCEL_EXPIRED = 4;         // Order has already expired
+  uint8 constant ERROR_CANCEL_NO_VALUE = 5;        // Order has already been fully filled or cancelled
 
   address public PROTOCOL_TOKEN;
   address public PROXY;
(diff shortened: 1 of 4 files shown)


contracts/Exchange.sol
```diff
@@ -202,8 +225,18 @@ contract Exchange is SafeMath {
...
+
+    if (block.timestamp >= expiration) {
+      LogError(ERROR_EXPIRED, orderHash);
+      return 0;
+    }
+
     cancelledValueM = min(cancelValueM, safeSub(values[0], fills[orderHash]));
-    if (cancelledValueM == 0) return 0;
+    if (cancelledValueM == 0) {
+      LogError(ERROR_FILLED_CANCELLED, orderHash);
```
fabioberger: Could we possibly have two separate error codes for filled and cancelled?

willwarren89: Unless we change how Exchange keeps track of fills/cancels it isn't possible to differentiate between the two. Cancelling an order sets `fills[orderHash] = valueM` where `valueM` is 100% of the order's value.

abandeali1: While we can't differentiate on the smart contract level right now, I think it does make sense to have 2 separate error codes for dApp developers (since we can differentiate within the fill/cancel functions).

I also think that we do want both `fills[orderHash` and `cancels[orderHash]` down the line. By doing this, smart contracts can verify that an order has been filled (or cancelled), which opens up some cool possibilities.

contracts/Exchange.sol
```diff
@@ -61,6 +67,8 @@ contract Exchange is SafeMath {
     bytes32 orderHash
   );
 
+  event LogError(uint8 indexed errorId, bytes32 orderHash);
```
abandeali1: Let's index `orderHash` as well, so developers can easily watch for specific orders failing.

contracts/Exchange.sol
abandeali1: I the error code names should be in the format `ERROR_{function_name}_{error_name}`, so we would actually have 2 errors for expired fills/cancels also.

Let's also add public getter functions that return the string versions of the error codes. It could make life a little easier for developers.

contracts/Exchange.sol
```diff
@@ -26,10 +26,12 @@ import "./base/SafeMath.sol";
...
   // Error Codes
-  uint8 constant ERROR_EXPIRED = 1;           // Order has already expired
-  uint8 constant ERROR_FILLED_CANCELLED = 2;  // Order has already been filled or cancelled
-  uint8 constant ERROR_TRUNCATION = 3;        // Rounding error too large
-  uint8 constant ERROR_BALANCE_ALLOWANCE = 4; // Insufficient balance or allowance
+  uint8 constant ERROR_FILL_EXPIRED = 1;           // Order has already expired
+  uint8 constant ERROR_FILL_NO_VALUE = 2;          // Order has already been fully filled or cancelled
+  uint8 constant ERROR_FILL_TRUNCATION = 3;        // Rounding error too large
+  uint8 constant ERROR_FILL_BALANCE_ALLOWANCE = 4; // Insufficient balance or allowance for token transfer
+  uint8 constant ERROR_CANCEL_EXPIRED = 5;         // Order has already expired
```
willwarren89: Does having FILL_EXPIRED and CANCEL_EXPIRED as separate error codes add any value here? Seems like the same error.

abandeali1: It's much better for analysis of historical data, and doesn't really add any overhead. I think the more specific with error codes, the better.

willwarren89: If I am understanding correctly, if someone is looking through past events they don't see the function that fired the event? In which case they would just see EXPIRED, but not know whether someone was attempting to fill or cancel an order?

abandeali1:
I'm not seeing much of a downside to inclusing these functions. Anything that makes it easier to programmatically interact with the smart contracts is a plus in my eyes, even if the functions aren't used internally. It probably makes sense to have a separate "Web3 call functions" section for these functions, though.

willwarren89:
How do you envision people using `getAllErrors` and `getError`?

contracts/Exchange.sol
abandeali1: They can see the function that fired the event if they look at the calldata of the txHash, but can't filter for it specifically. Having a single EXPIRED error would add an extra step and also make it a pain to filter for more specific errors in realtime.

abandeali1:
The same way I used `getAllErrCodes` in the tests, for example. I can see a case for removing `getErrCode` though, since `getAllErrCodes[errorId - 1]` will get the same result.

fabioberger:
I'd have to agree with @willwarren89 on this one. Let's not add the `getErrorCodes ` and instead have the listed error codes in the documentation.

PR #42, opened 2017-05-03
Change fillValueM => fillValueT

abandeali1:
This PR changes all instances of `fillValueM` to `fillValueT`. It is more intuitive for a taker to specify the amount of the token they wish to sell (since they already have that token), rather than the amount of the token they wish to buy. In addition, this change makes `fillUpTo` work much more similarly to a traditional market order.


willwarren89:
Can you also rename PROTOCOL_TOKEN to ZRX? This should improve code clarity.

PR #43, opened 2017-05-04
Add salt to order params

abandeali1:
This PR adds a salt to the order parameters. This is necessary to prevent traders from accidentally creating duplicate orders (with the same `orderHash`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ tokenT | `address` | Address of an ERC20 Token contract.
 valueM | `uint256` | Total units of tokenM offered by maker.
 valueT | `uint256` | Total units of tokenT requested by maker.
 expiration | `uint256` | Time at which the order expires (seconds since unix epoch).
-salt | `uint256` | Arbitrary number.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -44,6 +44,7 @@ tokenT | `address` | Address of an ERC20 Token contract.
 valueM | `uint256` | Total units of tokenM offered by maker.
 valueT | `uint256` | Total units of tokenT requested by maker.
 expiration | `uint256` | Time at which the order expires (seconds since unix epoch).
+salt | `uint256` | Arbitrary number.
```
fabioberger: Perhaps we could flesh out this explanation with:

```
Arbitrary number that guarantees orderHash uniqueness
```

Thoughts?

abandeali1: I don't like the word "guarantee" here, since someone could technically reuse the same salt. Maybe "supplement"?

fabioberger: How about:

``` 
Arbitrary number that allows for orderHash uniqueness
```

PR #44, opened 2017-05-08
Fix SoliditySHA3 encoding & Move amount types to BigNumber

fabioberger:
This PR:
- Replaces hard-coded '0x0' with constants.NULL_ADDRESS or constants.NULL_HASH as appropriate.
- Refactored amount values that are of type uint256 in solidity to type BigNumber at rest.
- Refactored amount values that are of type uint8 in solidity to type number.
- Fixed crypto.solSHA3 so that it properly encodes amount values that require more then 32 bits. The previous implementation limited us to 32 bit values while the contracts support up to 256 bits.
- Modified the salt param so that it is a very large random 256 bit number.
- Bonus: protected our codebase from the [2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem) by making our expirations BigNumbers instead of number. :1st_place_medal:


abandeali1:
Looks good!

PR #46, opened 2017-05-10
MultiSigWalletWithTimeLock

abandeali1:
This PR:

-Creates a contract `MultiSigWalletWithTimeLock`, which inherits the original Consensys MultiSig found here: https://github.com/ConsenSys/MultiSigWallet/blob/master/contracts/solidity/MultiSigWallet.sol
This should drastically improve readability and make it easier to differentiate between the 2 contracts.
-FIxes tests to work with new variable names and slight interface changes

### DIFF
--- a/contracts/MultiSigWalletWithTimeLock.sol
+++ b/contracts/MultiSigWalletWithTimeLock.sol
@@ -36,9 +36,9 @@ contract MultiSigWalletWithTimeLock is MultiSigWallet {
     }
 
     modifier pastTimeLock(uint transactionId) {
-      uint confirmationTime = confirmationTimes[transactionId];
-      if (confirmationTime == 0 || block.timestamp < confirmationTime + secondsTimeLocked) throw;
-      _;
+        uint confirmationTime = confirmationTimes[transactionId];
+        if (confirmationTime == 0 || block.timestamp < confirmationTime + secondsTimeLocked) throw;
+        _;
     }
 
     /*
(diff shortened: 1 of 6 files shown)


contracts/MultiSigWalletWithTimeLock.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    uint public secondsTimeLocked;
+    mapping (uint => uint) public confirmationTimes;
+
+    modifier confirmationTimeNotSet(uint transactionId) {
+        if (confirmationTimes[transactionId] != 0) throw;
+        _;
+    }
+
+    modifier pastTimeLock(uint transactionId) {
```
fabioberger: It looks to me like this modifier is checking two conditions:
- Whether the confirmationTime is set
- Whether the timeLock has past

Whereas the modifiers name only  alludes to the second condition. 

Let's either make another modifier called `confirmationTimeSet` or rename this modifier to `confirmationTimeSetAndPastTimeLock`. I prefer the first option.

test/ts/multi_sig.ts
```diff
@@ -65,52 +65,52 @@ contract('MultiSigWallet', (accounts: string[]) => {
...
+      assert.equal(initialSecondsTimeLocked, 0);
 
       const res = await multiSig.executeTransaction(txId);
       assert.equal(res.logs.length, 2);
 
-      const threshold = await multiSig.secondsRequired.call();
-      const newThreshold = threshold.toNumber();
-      assert.equal(newThreshold, SECONDS_REQUIRED);
+      const secondsTimeLocked = await multiSig.secondsTimeLocked.call();
+      const newSecondsTimeLocked = secondsTimeLocked.toNumber();
```
fabioberger: Let's consistently convert the timeLock BigNumber to the same JS value for comparisons. Above we compared them as strings, and here as numbers. For now, I'd pick strings since this is consistent with the rest of the code base.

abandeali1: I decided to pick numbers. Feels very awkward doing `assert.equal(..., "0")`, and some of the outputs were already numbers.

PR #48, opened 2017-05-10
SimpleCrowdsale

abandeali1:
This PR:

-Adds a contract `SimpleCrowdsale` that can be used to launch a capped crowdsale. Users can send ETH to the contract to fill the order, or manually fill the order using the `Exchange` contract.

-Replaces `DummyEtherToken` with an updated `EtherToken`. This is the same contract found here, except it is modified to use our current version of `SafeMath`: https://github.com/ConsenSys/gnosis-contracts/blob/master/contracts/solidity/Tokens/EtherToken.sol


LogvinovLeon:
Two questions:
* Why do we merge variables in two-element arrays? `rs`, `expirationAndSalt`, `tokens`? 
Why shouldn't we have two separate variables with two separate descriptive names? Gas cost?
* I'm not a big fan of tests that rely on the ordering of tests. It's easy to introduce an unintended side-effect then or even to have a situation, where you can not add the test you want. For example in a current approach we can have only one test that initializes the Crowdsale. Why don't we use [truffle's approach](http://truffleframework.com/docs/getting_started/testing#clean-room-environment)? Then we would also not need to reimplement migrations in tests.

LogvinovLeon:
For now (mocha doesn't support)[https://github.com/mochajs/mocha/issues/902] random test ordering, cause they think it's easy to check yourself and not introduce side-effects without tooling support, but having an experience of working with a codebase with tens of thousands of tests I can say, that those unintended side-effects are hard to find and are causing most of the spurious test failures later.

fabioberger:
Both issues you mentioned we've already talked about. Re: two-element array param, we've hit the stack limit on number of function params for Solidity. This is purely a hack to add more args. There is currently no way around this :disappointed:

fabioberger:
As for test ordering dependencies, that is in my mind the #1 refactor that needs to happen to our contract tests. It's only a matter of time before we do it. 100% agree with your issue with it :slightly_smiling_face:

migrations/3_register_tokens.ts
```diff
@@ -25,7 +25,7 @@ module.exports = (deployer: any, network: string) => {
         totalSupply,
     )), _.noop).then((dummyTokens: ContractInstance[]) => {
         const weth = {
-          address: DummyEtherToken.address,
+          address: EtherToken.address,
           name: 'ETH Token',
```
fabioberger: Let's make sure this name is the same as in the smart contract to avoid confusion: `Ether Token`

migrations/5_configure_crowdsale.ts
```diff
@@ -0,0 +1,24 @@
...
+const {
+  Proxy,
+  Exchange,
+  TokenRegistry,
+  SimpleCrowdsale,
+} = new Artifacts(artifacts);
+
+let tokenRegistry: ContractInstance;
+module.exports = (deployer: any) => {
+  deployer.then(() => TokenRegistry.deployed()
```
fabioberger: Let's add brackets around this method since although technically a one-liner, it spans multiple lines. This will help with the readability of this logic.

migrations/5_configure_crowdsale.ts
```diff
@@ -0,0 +1,24 @@
...
+  .then((instance: ContractInstance) => {
+    tokenRegistry = instance;
+    return Promise.all([
```
fabioberger: Let's use Bluebird's`Bluebird.each` so that this is guaranteed to work on the testnet/livenet.

test/ts/simple_crowdsale.ts
```diff
@@ -0,0 +1,255 @@
...
+          [order.params.r, order.params.s],
+          { from: notOwner },
+        );
+        throw new Error('Init succeeded when it should have thrown');
+      } catch (err) {
+        testUtil.assertThrow(err);
+      }
+    });
+
+    it('should throw if called with an invalid order', async () => {
```
fabioberger: Let's also add a test for if `init` called with invalidTokens.

test/ts/simple_crowdsale.ts
```diff
@@ -0,0 +1,255 @@
...
+        value: ethValueSent,
+        gas: 300000,
+        gasPrice,
+      });
+      const receipt = await getTransactionReceipt(txHash);
+
+      const finalBalances: BalancesByOwner = await dmyBalances.getAsync();
+      const finalTakerEthBalance = await getEthBalance(taker);
+      const ethSpentOnGas = mul(receipt.gasUsed, gasPrice);
+      const zrxValue = cmp(expectedZrxValue, remainingValueT) > 0 ? remainingValueT : expectedZrxValue;
```
fabioberger: When would `zrxValue` be equal to `remainingValueT ` and when would it be equal to `expectedZrxValue `? Shouldn't this test run deterministically and always be equal to one or the other?

test/ts/simple_crowdsale.ts
```diff
@@ -0,0 +1,255 @@
...
+          value: ethValue,
+          gas: 300000,
+        });
+        throw new Error('Fallback succeeded when it should have thrown');
+      } catch (err) {
+        testUtil.assertThrow(err);
+      }
+    });
+  });
+});
```
fabioberger: Missing test: it should throw if someone sends the contract ETH after the order has expired.

migrations/5_configure_crowdsale.ts
abandeali1: Is there any reason we aren't already using Bluebird.each in steps `configure_proxy` and `transfer_ownership`?

test/ts/simple_crowdsale.ts
abandeali1: I've actually been thinking about removing that check anyways, since `fillOrKill` will throw if the order is expired anyways. Thoughts?

test/ts/simple_crowdsale.ts
abandeali1: If someone tries to fill > than the remaining amount, the contract will instead only fill the remaining amount. This line is using the same logic to assign a value to `zrxValue`.

migrations/5_configure_crowdsale.ts
fabioberger: From triaging Truffle race-conditions it seems like they occur when attempting to run multiple calls that interact with the blockchain (contract deploys & method calls). In `configure_proxy ` and `transfer_ownership ` we are simply getting an instance of an already deployed contract, which in theory should only require retrieving information from the artifacts. Hence, they do not seem to elicit race conditions.

test/ts/simple_crowdsale.ts
fabioberger: Each test has a starting state and an end state. Given the start state is always the same, shouldn't we know (and hence be able to "hard-code") the expected zrxValue?

Is this logic trying to future-proof this test to people changing previous tests (hence changing the start state for this test)? If so, we should remove this logic because if that were to happen, this test's description would no longer accurately describe this test and would need to be refactored.

test/ts/simple_crowdsale.ts
fabioberger: That makes sense to me! Let's remove it.

PR #51, opened 2017-05-15
Refactor Exchange params

abandeali1:
This PR:

-Adds an `Order` struct to the `Exchange` contract.
-Removes all of the tuple params from all of the `Exchange` methods in favor of `orderAddresses` and `orderValues` arrays.

With the above changes, we free up local variable slots (we were previously at the max) and increase code readability. The new local variable slots allow us to make the below changes:

-Remove `fillSuccess` and `cancelSuccess`.
-The `rs` param has been separated into `r` and `s` (I could still see this going either way).
-Filled and cancelled amounts are now stored in two separate mappings, `filled` and `cancelled`. It could be useful to have on-chain proof that specific orders were either filled or cancelled, where it was previously impossible to differentiate. A helper method `getUnavailableValueT` has also been added, which returns the sum of `filled[orderHash]` and `cancelled[orderHash]`.

Other minor changes:
-Private methods are now internal.

### DIFF
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -49,6 +49,7 @@ contract Exchange is SafeMath {
         uint feeM,
         uint feeT,
         uint expiration,
+        uint salt,
         uint filledValueT,
         bytes32 indexed tokens,
         bytes32 orderHash
@@ -64,12 +65,28 @@ contract Exchange is SafeMath {
         uint feeM,
         uint feeT,
         uint expiration,
+        uint salt,
         uint cancelledValueT,
         bytes32 indexed tokens,
         bytes32 orderHash
     );
 
     event LogError(uint8 indexed errorId, bytes32 indexed orderHash);
+
+    struct Order {
+        address maker;
+        address taker;
+        address tokenM;
+        address tokenT;
+        address feeRecipient;
+        uint valueM;
+        uint valueT;
+        uint feeM;
+        uint feeT;
+        uint expiration;
+        uint salt;
+        bytes32 orderHash;
+    }
 
     function Exchange(address _zrx, address _proxy) {
         ZRX = _zrx;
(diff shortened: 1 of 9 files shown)


contracts/Exchange.sol
```diff
@@ -81,176 +98,177 @@ contract Exchange is SafeMath {
...
         assert(isValidSignature(
-            traders[0],
-            orderHash,
+            order.maker,
+            order.orderHash,
             v,
-            rs[0],
-            rs[1]
+            r,
+            r
```
willwarren89: This should be `s`?

abandeali1: Yep, that was fixed in commit 	070cc96.

contracts/Exchange.sol
```diff
@@ -461,36 +427,26 @@ contract Exchange is SafeMath {
...
+            orderAddresses[1],
+            orderAddresses[2],
+            orderAddresses[3],
+            orderAddresses[4],
+            orderValues[0],
+            orderValues[1],
+            orderValues[2],
+            orderValues[3],
+            orderValues[4],
+            orderValues[5]
```
fabioberger: Since this has become less readable with this refactor, let's add comments next to each of these with the value it represents. E.g

```
orderAddresses[0], /// maker
orderAddresses[1], /// taker
orderAddresses[2], /// tokenM
```

contracts/SimpleCrowdsale.sol
```diff
@@ -158,36 +152,26 @@ contract SimpleCrowdsale is Ownable, SafeMath {
...
+            orderAddresses[1],
+            orderAddresses[2],
+            orderAddresses[3],
+            orderAddresses[4],
+            orderValues[0],
+            orderValues[1],
+            orderValues[2],
+            orderValues[3],
+            orderValues[4],
+            orderValues[5]
```
fabioberger: Ditto to comment above about adding comments next to each one of these.

PR #61, opened 2017-06-16
Fix isTransferable

abandeali1:
-Fixes isTransferable potentially returning the wrong result when one of the order tokens is ZRX

### DIFF
--- a/test/ts/exchange/core.ts
+++ b/test/ts/exchange/core.ts
@@ -345,7 +345,8 @@ contract('Exchange', (accounts: string[]) => {
       }
     });
 
-    it('should not change balances if maker balances are too low to fill order and shouldCheckTransfer = true', async () => {
+    it('should not change balances if maker balances are too low to fill order and shouldCheckTransfer = true',
+       async () => {
       order = await orderFactory.newSignedOrderAsync({
         valueM: toSmallestUnits(100000),
       });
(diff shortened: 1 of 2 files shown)


willwarren89:
Should we update variable names as per [our quip doc](https://quip.com/p54OAhSKsjZV)?

abandeali1:
Have another PR coming up with variable renames and upgrade to Solidity 0.4.11

test/ts/exchange/core.ts
```diff
@@ -422,6 +424,60 @@ contract('Exchange', (accounts: string[]) => {
...
+      });
+      await exWrapper.fillAsync(order, taker, { shouldCheckTransfer: true });
+      const newBalances = await dmyBalances.getAsync();
+      assert.deepEqual(newBalances, balances);
+    });
+
+    it('should not change balances if tokenM is ZRX, valueM + feeM > maker allowance, and shouldCheckTransfer = true',
+       async () => {
+      const makerZRXAllowance = await zrx.allowance(maker, Proxy.address);
+      const makerZRXAllowanceFixed = makerZRXAllowance.toFixed(); // remove scientific notation
```
fabioberger: This is a hack. We should configure BigNumber to not return scientific notation by setting the 
EXPONENTIAL_AT param to 1000. See `bignumber_config.ts` in `0x.js`.

Also, the version of BigNumber returned from Web3 is very outdated. We should always re-wrap
BigNumbers right after retrieval. This will is also the only way to set our own configs.

PR #62, opened 2017-06-16
Naming conventions + upgrade Solidity

abandeali1:
This PR:
-Upgrades Solidity to 0.4.11 (requires truffle ^3.2.2)
-Makes use of require/assert instead of most instances of `if(...) throw`
-Uses the following new naming conventions for contracts:

* batchFill → batchFillOrders
* fillUpTo → fillOrdersUpTo
* batchCancel → batchCancelOrders
* shouldCheckTransfer → shouldThrowOnInsufficientBalanceOrAllowance
* tokenM → makerToken
* tokenT → takerToken
* valueM → makerTokenAmount
* valueT → takerTokenAmount
* feeT → takerFee
* feeM → makerFee
* feeMPaid → paidMakerFee
* feeTPaid → paidTakerFee
* fillValueM → fillMakerTokenAmount
* fillValueT → fillTakerTokenAmount
* filledValueM → filledMakerTokenAmount
* filledValueT → filledTakerTokenAmount
* cancelledValueM → cancelledMakerTokenAmount
* cancelledValueT → cancelledTakerTokenAmount
* remainingValueM → remainingMakerTokenAmount
* remainingValueT → remainingTakerTokenAmount
* expiration → expirationTimestampInSec
* ZRX → ZRX_TOKEN_CONTRACT
* PROXY → PROXY_CONTRACT
* ERROR_FILL_EXPIRED → ERROR_ORDER_EXPIRED
* ERROR_FILL_NO_VALUE → ERROR_ORDER_FULLY_FILLED_OR_CANCELLED
* ERROR_FILL_TRUNCATION → ERROR_ROUNDING_ERROR_TOO_LARGE
* ERROR_FILL_BALANCE_ALLOWANCE → ERROR_INSUFFICIENT_BALANCE_OR_ALLOWANCE
* ERROR_CANCEL_EXPIRED → ERROR_ORDER_EXPIRED
* ERROR_CANCEL_NO_VALUE → ERROR_ORDER_FULLY_FILLED_OR_CANCELLED


util/test_util.ts
```diff
@@ -4,7 +4,7 @@ import * as assert from 'assert';
 export const testUtil = {
   assertThrow(err: Error) {
     const errMsg = `${err}`;
-    const didCalledContractThrow = _.includes(errMsg, 'invalid JUMP');
+    const didCalledContractThrow = _.includes(errMsg, 'invalid opcode');
```
fabioberger: From the Solidity release notes:

```
Internal exceptions are now thrown by using an invalid opcode (0xfe), manual exceptions still use an invalid jump.
```

Shouldn't we check for both `invalid JUMP` **and** `invalid opcode`? I'm not entirely sure I understand what is meant by "manual exceptions".

contracts/base/StandardToken.sol
```diff
@@ -1,12 +1,12 @@
...
+pragma solidity ^0.4.11;
 
 import "./Token.sol";
 
 contract StandardToken is Token {
 
     function transfer(address _to, uint _value) returns (bool success) {
         //Default assumes totalSupply can't be over max (2^256 - 1).
-        if (balances[msg.sender] >= _value && balances[_to] + _value > balances[_to]) {
+        if (balances[msg.sender] >= _value && balances[_to] + _value >= balances[_to]) {
```
abandeali1: It seems to be general consensus that a Standard Token should not fail on a 0-value transfer. I wouldn't say it's a bugfix, but our Standard Token was a bit outdated.

test/ts/exchange/core.ts
```diff
@@ -56,15 +56,15 @@ contract('Exchange', (accounts: string[]) => {
     ]);
 
     const defaultOrderParams = {
-      exchange: Exchange.address,
+      version: Exchange.address,
```
abandeali1: For consistency with the whitepaper/README, but I am also open to changing this.

contracts/base/SafeMath.sol
```diff
@@ -1,32 +1,28 @@
...
+        require(c >= a && c >= b);
         return c;
     }
 
     function safeDiv(uint a, uint b) constant internal returns (uint) {
-        assert(b > 0);
+        require(b > 0);
         uint c = a / b;
-        assert(a == b * c + a % b);
+        require(a == b * c + a % b);
```
abandeali1: Where is the line drawn for require/assert? If the code was instead `require(a == b * (a / b) + a % b);`, would that still be considered an internal error?

contracts/Proxy.sol
```diff
@@ -98,8 +98,7 @@ contract Proxy is Ownable {
         onlyAuthorized
         returns (bool success)
     {
-        if (!Token(token).transferFrom(from, to, value)) throw;
-        return true;
+        return Token(token).transferFrom(from, to, value);
```
abandeali1: This pretty much sums my thoughts up: https://github.com/ethereum/EIPs/pull/610#issuecomment-305770167

I don't see much point in limiting its functionality when its much easier to force a `return false` to `throw` than to do the reverse. There might not be amazing examples of a graceful failure on transfer being necessary today, but if we expect more and more complex systems of smart contracts to be built over time then I'm sure the situation will arise more frequently.

util/test_util.ts
abandeali1: I think you are right, we should be checking for both. From the Solidity docs:

`Internally, Solidity performs a revert operation (instruction 0xfd) when a user-provided exception is thrown or the condition of a require call is not met. In contrast, it performs an invalid operation (instruction 0xfe) if a runtime exception is encountered or the condition of an assert call is not met.`

contracts/base/SafeMath.sol
fabioberger: Talked offline about this.

contracts/base/StandardToken.sol
fabioberger: I see. Adding a test for this could still be good.

test/ts/exchange/core.ts
fabioberger: I do feel like this is less intuitive. If I saw that I had to supply the version, I'd have to do some research to understand what versioning system 0x uses. If it says `exchangeContractAddress`, I know exactly what I need to supply.

PR #66, opened 2017-06-23
Add initial exchange_migration script

LogvinovLeon:
This PR:
* Adds a cli script that gnerates the call data to be submited to a multisig in order to update Exchange contract
* Is probably not needed, cause I checked gnosis UI after doing this and they also have this functionality.
You can submit the contract abi and call arbitrary function.
I was not able to check that cause I didn't have a private key to any of the authorised addreses
screen shot 2017-06-23 at 14 24 03


abandeali1:
The issue with the UI is that it has no way of executing a submission once fully confirmed. In the original MultiSigWallet implementation, `execute` is called automatically so this isn't an issue. Since we have a timelock, we have to call `execute` manually after enough time has passed.

So it would be nice if a) submitting a tx outputs the txId from the MultiSig and b) we have a script for executing a txId.

PR #67, opened 2017-06-25
Add fallback to EtherToken

abandeali1:
Adds a fallback to EtherToken which calls `deposit`.

### DIFF
--- a/contracts/tokens/EtherToken.sol
+++ b/contracts/tokens/EtherToken.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.8;
+pragma solidity ^0.4.11;
 
 import "../base/StandardTokenWithOverflowProtection.sol";
 
@@ -17,9 +17,18 @@ contract EtherToken is StandardTokenWithOverflowProtection {
     /*
      *  Read and write functions
      */
+
+    /// @dev Send ether to contract to deposit.
+    function()
+        public
+        payable
+    {
+        deposit();
+    }
+
     /// @dev Buys tokens with Ether, exchanging them 1:1.
     function deposit()
-        external
+        public
         payable
     {
         balances[msg.sender] = safeAdd(balances[msg.sender], msg.value);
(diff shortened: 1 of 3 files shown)


contracts/tokens/EtherToken.sol
```diff
@@ -17,9 +17,18 @@ contract EtherToken is StandardTokenWithOverflowProtection {
     /*
      *  Read and write functions
      */
+
+    /// @dev Send ether to contract to deposit.
```
fabioberger: How about: `Fallback to calling deposit when ether is sent directly to contract`

contracts/tokens/EtherToken.sol
```diff
@@ -29,13 +38,10 @@ contract EtherToken is StandardTokenWithOverflowProtection {
     /// @dev Sells tokens in exchange for Ether, exchanging them 1:1.
     /// @param amount Number of tokens to sell.
     function withdraw(uint amount)
-        external
+        public
```
fabioberger: This could still be `external`. Any reason to switch it to `public`?

abandeali1: Mostly for consistency with other contracts, we haven't really made use of any external methods.

test/ts/ether_token.ts
```diff
@@ -0,0 +1,108 @@
...
+      } catch (err) {
+        testUtil.assertThrow(err);
+      }
+    });
+
+    it('should convert ether tokens to ether with sufficient balance', async () => {
+      const initEthTokenBalance = await ethToken.balanceOf(account);
+      const initEthBalance = await getEthBalance(account);
+      const ethTokensToWithdraw = initEthTokenBalance;
+      assert.equal(cmp(ethTokensToWithdraw, 0), 1);
```
fabioberger: This is highly unreadable... Let's do:

```
assert.equal(ethTokensToWithdraw.toNumber(), 0);
```

test/ts/ether_token.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      const gasPrice = new BigNumber(web3Instance.toWei(20, 'gwei'));
```
fabioberger: This constant is used in multiple tests. Let's declare it in the `contract()` scope above. This way we only declare it in one place.

PR #68, opened 2017-06-25
Add CrowdsaleWithRegistry

abandeali1:
This PR:
-Adds a per person cap to the sample crowdsale contract that only allows registered addresses to participate.
-Adds events to SimpleCrowdsale.


test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+        feeRecipient: constants.NULL_ADDRESS,
+        makerToken: zrxAddress,
+        takerToken: wEthAddress,
+        makerTokenAmount: toSmallestUnits(5),
+        takerTokenAmount: toSmallestUnits(5),
+        makerFee: new BigNumber(0),
+        takerFee: new BigNumber(0),
+        expirationTimestampInSec: new BigNumber(Math.floor(Date.now() / 1000) + 1000000000),
+        salt: new BigNumber(0),
+      };
```
fabioberger: This is identical to the `validOrderParams` declared further up in the file. Any reason we cannot declare it once and share it between tests?

test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+        { from: owner },
+      );
+      await crowdsaleWithRegistry.registerAddress(taker, { from: owner });
+      try {
+        const ethValue = new BigNumber(1);
+        await crowdsaleWithRegistry.fillOrderWithEth({
+          from: taker,
+          value: ethValue,
+        });
+        throw new Error('Fallback succeeded when it should have thrown');
```
fabioberger: This error message does not pertain to the test case.

test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+      const initTakerEthBalance = await getEthBalance(taker);
+
+      const ethValue = web3Instance.toWei(1, 'ether');
+      const zrxValue = div(mul(ethValue, validOrder.params.makerTokenAmount), validOrder.params.takerTokenAmount);
+      const gasPrice = web3Instance.toWei(20, 'gwei');
+
+      const res = await crowdsaleWithRegistry.fillOrderWithEth({
+        from: taker,
+        value: ethValue,
+        gas: 300000,
```
fabioberger: Does this call not work without specifying gas explicitly? If so, how did you reach this number? It would be better if we called `estimateGas` instead of hard-coding this amount here. (e.g https://github.com/0xProject/0x.js/blob/master/src/contract_wrappers/exchange_wrapper.ts#L160)

test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+      }
+    });
+
+    it('should trade sent ETH for protocol tokens if sender is registered, ETH <= remaining order ETH and capPerAddress', async () => {
+      const initBalances: BalancesByOwner = await dmyBalances.getAsync();
+      const initTakerEthBalance = await getEthBalance(taker);
+
+      const ethValue = web3Instance.toWei(1, 'ether');
+      const zrxValue = div(mul(ethValue, validOrder.params.makerTokenAmount), validOrder.params.takerTokenAmount);
+      const gasPrice = web3Instance.toWei(20, 'gwei');
```
fabioberger: This should be defined higher up since it is a constant used in multiple places.

test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+       async () => {
+      const initBalances: BalancesByOwner = await dmyBalances.getAsync();
+      const initTakerEthBalance = await getEthBalance(taker);
+      const ethValue = add(capPerAddress, 1);
+      const gasPrice = web3Instance.toWei(20, 'gwei');
+
+      const res = await crowdsaleWithRegistry.fillOrderWithEth({
+        from: taker,
+        value: ethValue,
+        gas: 300000,
```
fabioberger: This magic number needs to be assigned to a constant and have a descriptive name.

test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+    it('should partial fill and end sale if sender is registered and sent ETH > remaining order ETH', async () => {
+      const newCapPerAddress = mul(validOrder.params.makerTokenAmount, 2);
+      await crowdsaleWithRegistry.setCapPerAddress(newCapPerAddress, { from: owner });
+      const initBalances: BalancesByOwner = await dmyBalances.getAsync();
+      const initTakerEthBalance = await getEthBalance(taker);
+      const remainingTakerTokenAmount = sub(validOrder.params.takerTokenAmount,
+                                            await exchange.getUnavailableTakerTokenAmount(validOrder.params.orderHashHex));
+
+      const ethValue = web3Instance.toWei(6, 'ether');
+      const gasPrice = web3Instance.toWei(6, 'gwei');
```
fabioberger: Why is the gasPrice 6 gwei here and 20 gwei everywhere else?

test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+
+      assert.equal(finalBalances[maker][validOrder.params.makerToken],
+                   sub(initBalances[maker][validOrder.params.makerToken], filledZrxValue));
+      assert.equal(finalBalances[maker][validOrder.params.takerToken],
+                   add(initBalances[maker][validOrder.params.takerToken], filledEthValue));
+      assert.equal(finalBalances[taker][validOrder.params.makerToken],
+                   add(initBalances[taker][validOrder.params.makerToken], filledZrxValue));
+      assert.equal(finalTakerEthBalance, sub(sub(initTakerEthBalance, filledEthValue), ethSpentOnGas));
+
+      assert.equal(res.receipt.logs.length, 5);
```
fabioberger: What are the 5 events you expect to be fired?

test/ts/crowdsale_with_registry.ts
```diff
@@ -0,0 +1,646 @@
...
+          gas: 300000,
+        });
+        throw new Error('Fallback succeeded when it should have thrown');
+      } catch (err) {
+        testUtil.assertThrow(err);
+      }
+    });
+  });
+
+  describe('fallback', () => {
```
fabioberger: This describe block has the exact same `beforeEach` as the previous describe block. Why not create a shared "parent" describe block with this `beforeEach` block and then nest these two function-named describe block within it? That would reduce code duplication.

LogvinovLeon:
Contracts are OK.
Tests have some duplication and can benefit from using 0x.js and bignumber assertions, but theis is not the scope of this PR, so after fixing Fabio's comments - this is good to go for me.

test/ts/crowdsale_with_registry.ts
abandeali1: We have to reassign taker to the new `crowdsaleWithRegistry` address.

test/ts/crowdsale_with_registry.ts
abandeali1: Oops, typo.

fabioberger: And the tests still passed?

test/ts/token_distribution_with_registry.ts
```diff
@@ -0,0 +1,539 @@
...
+        const initBalances: BalancesByOwner = await dmyBalances.getAsync();
+        const initTakerEthBalance = await getEthBalanceAsync(taker);
+
+        const ethValue = web3Instance.toWei(1, 'ether');
+        const zrxValue = div(mul(ethValue, validOrder.params.makerTokenAmount), validOrder.params.takerTokenAmount);
+
+        const res = await tokenDistributionWithRegistry.fillOrderWithEth({
+          from: taker,
+          value: ethValue,
+          gasPrice,
```
fabioberger: Aren't you supposed to specify the gas amount here, not the price?

test/ts/token_distribution_with_registry.ts
```diff
@@ -0,0 +1,539 @@
...
+        await tokenDistributionWithRegistry.setCapPerAddress(newCapPerAddress, { from: owner });
+        const initBalances: BalancesByOwner = await dmyBalances.getAsync();
+        const initTakerEthBalance = await getEthBalanceAsync(taker);
+
+        const ethValue = add(validOrder.params.takerTokenAmount, 1);
+
+        const res = await tokenDistributionWithRegistry.fillOrderWithEth({
+          from: taker,
+          value: ethValue,
+          gasPrice,
```
fabioberger: Ditto comment above.

test/ts/token_distribution_with_registry.ts
```diff
@@ -0,0 +1,539 @@
...
+      it('should throw if sale has ended', async () => {
+        const newCapPerAddress = mul(validOrder.params.makerTokenAmount, 2);
+        await tokenDistributionWithRegistry.setCapPerAddress(newCapPerAddress, { from: owner });
+
+        const ethValue = add(validOrder.params.takerTokenAmount, 1);
+
+        await tokenDistributionWithRegistry.fillOrderWithEth({
+          from: taker,
+          value: ethValue,
+          gasPrice,
```
fabioberger: Ditto above.

test/ts/token_distribution_with_registry.ts
```diff
@@ -0,0 +1,539 @@
...
+    describe('fallback', () => {
+      it('should trade sent ETH for protocol tokens if ETH <= remaining order ETH and ethCapPerAddress',
+         async () => {
+        const initBalances: BalancesByOwner = await dmyBalances.getAsync();
+        const initTakerEthBalance = await getEthBalanceAsync(taker);
+
+        const ethValue = web3Instance.toWei(1, 'ether');
+        const zrxValue = div(mul(ethValue, validOrder.params.makerTokenAmount), validOrder.params.takerTokenAmount);
+
+        const gas = 300000;
```
fabioberger: Where did this number come from? Can you add a comment above this explaining where this came from?

test/ts/token_distribution_with_registry.ts
abandeali1: We specify the price so that we can predict the taker's expected ETH balance after the function is called. We only need to specify the gas amount when testing the fallback, since the default gas it sends for that is too low.

PR #70, opened 2017-06-28
Add MultiSigWithTimeLockExceptRemoveAuthorizedAddress

abandeali1:
Adds a contract `MultiSigWithTimeLockExceptRemoveAuthorizedAddress`, which is a `MultiSigWalletWithTimeLock` but can call `proxy.removeAuthorizedAddress` with no delay.


contracts/ProxyOwner.sol
```diff
@@ -0,0 +1,54 @@
+pragma solidity ^0.4.11;
+
+import "./MultiSigWalletWithTimeLock.sol";
+
+contract ProxyOwner is MultiSigWalletWithTimeLock {
```
fabioberger: Why is this called `proxyOwner`? It doesn't describe the `multiSigWallet` aspect of the contract. Perhaps a better name would be: `MultiSigWithTimeLockExceptRemoveAuthorizedAddress`? I know it's super verbose but it's at least clear from the title exactly what the smart contract does.

contracts/ProxyOwner.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    address public PROXY_CONTRACT;
+
+    modifier validTxData(uint transactionId) {
```
fabioberger: Let's rename this to `validRemoveAuthorizedAddressTx`. When I saw this modifier on the function below, I had no idea it was checked to make sure the txData was being checked for the `removeAuthorizedAddress` function signature.

contracts/ProxyOwner.sol
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    function executeRemoveAuthorizedAddress(uint transactionId)
+        public
+        notExecuted(transactionId)
+        confirmationTimeSet(transactionId)
+        validTxData(transactionId)
+    {
+        Transaction tx = transactions[transactionId];
+        tx.executed = true;
```
fabioberger: Shouldn't this line be within the subsequent `if` block? It should only set `executed` to true if the execution succeeds. I know this logic works currently, I just feel it would be cleaner to have this in the if block.

contracts/ProxyOwner.sol
```diff
@@ -0,0 +1,54 @@
...
+        constant
+        returns (bool)
+    {
+        bytes4 removeAuthorizedAddressSignature = bytes4(sha3("removeAuthorizedAddress(address)"));
+        for (uint i = 0; i < 4; i++) {
+            assert(data[i] == removeAuthorizedAddressSignature[i]);
+        }
+        return true;
+    }
+}
```
fabioberger: Nice! I like how clear the separate of this logic is from the `multiSigWalletWithTimelock`, great job!

test/ts/proxy_owner.ts
```diff
@@ -0,0 +1,153 @@
...
+import { ContractInstance, TransactionDataParams } from '../../util/types';
+import { testUtil } from '../../util/test_util';
+import * as proxyJSON from '../../build/contracts/Proxy.json';
+const { Proxy, ProxyOwner } = new Artifacts(artifacts);
+const PROXY_ABI = (proxyJSON as any).abi;
+
+contract('ProxyOwner', (accounts: string[]) => {
+  const owners = [accounts[0], accounts[1]];
+  const authorizedAddress = accounts[0];
+  const unauthorizedAddress = accounts[1];
```
fabioberger: Nit: It's a bit confusing to me that we are using user addresses for these examples when the actual addresses that are going to be authorized are contract addresses. Any way to get contract addresses or fake addresses instead?

test/ts/proxy_owner.ts
```diff
@@ -0,0 +1,153 @@
...
+  const required = 2;
```
fabioberger: Let's rename this variable to `requiredApprovals`.

contracts/ProxyOwner.sol
abandeali1: It's done this way to prevent reentrancy. If you move `tx.executed` into the `if` block, you can execute a tx which does something else and then calls back into execute on the same tx. It's probably pretty hard for this to happen in reality, but it's generally considered best practice to change state before making external function calls.

contracts/MultiSigWalletWithTimeLock.sol
```diff
@@ -48,7 +50,7 @@ contract MultiSigWalletWithTimeLock is MultiSigWallet {
      * Public functions
      */
 
-    /// @dev Contract constructor sets initial owners and required number of confirmations.
+    /// @dev Contract constructor sets initial owners, required number of confirmations, and time lock.
```
LogvinovLeon: Remove the coma

contracts/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -0,0 +1,64 @@
...
+    /// @param data Transaction data.
+    /// @return Successful if data is a call to removeAuthorizedAddress.
+    function isFunctionRemoveAuthorizedAddress(bytes data)
+        public
+        constant
+        returns (bool)
+    {
+        bytes4 removeAuthorizedAddressSignature = bytes4(sha3("removeAuthorizedAddress(address)"));
+        for (uint i = 0; i < 4; i++) {
+            assert(data[i] == removeAuthorizedAddressSignature[i]);
```
LogvinovLeon: Doesn't `bytes4(data) == removeAuthorizedAddressSignature` work?

abandeali1: You can't explicitly convert a dynamically sized byte array to a fixed size one.

PR #78, opened 2017-07-06
Fix the dependencies link in the README

elopio:


fabioberger:
Thanks for the fix @elopio! 👍

PR #79, opened 2017-07-07
Update truffle and testrpc, remove hashPersonalMessage from order signing

abandeali1:
-As of ethereumjs-testrpc v4.0.0, eth_sign automatically hashes the Ethereum Signed Message header with the message, so this step was removed from order signing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Install [Node v6.9.1](https://nodejs.org/en/download/releases/)
 Install ethereumjs-testrpc
 
 ```
-npm i -g ethereumjs-testrpc@^3.0.2
+npm i -g ethereumjs-testrpc@^4.0.1
 ```
 
 Install project dependencies:
--- a/README.md
+++ b/README.md
@@ -84,12 +84,6 @@ s | `bytes32` | ECDSA signature of the above arguments.
 
 Install [Node v6.9.1](https://nodejs.org/en/download/releases/)
 
-Install ethereumjs-testrpc
-
-```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -87,7 +87,7 @@ Install [Node v6.9.1](https://nodejs.org/en/download/releases/)
 Install ethereumjs-testrpc
 
 ```
-npm i -g ethereumjs-testrpc@^3.0.2
+npm i -g ethereumjs-testrpc@^4.0.1
```
LogvinovLeon: Please add testrpc a a dev dependency

circle.yml
```diff
@@ -4,12 +4,12 @@ machine:
 
 dependencies:
   pre:
-    - npm install -g npm@3.10.10
-    - npm install ethereumjs-testrpc@3.0.2
-    - npm install -g truffle@3.2.5
+    - npm install -g npm@5.1.0
+    - npm install ethereumjs-testrpc@4.0.1
```
LogvinovLeon: And this line would not be required any more

circle.yml
```diff
@@ -4,12 +4,12 @@ machine:
...
+    - npm install -g truffle@3.4.3
 
 test:
   override:
-    - node node_modules/ethereumjs-testrpc/bin/testrpc:
+    - node node_modules/ethereumjs-testrpc/build/cli.node.js:
```
LogvinovLeon: And use `npm run testrpc` here. (Add testrpc command to package.json)

Issue #75, opened 2017-07-05
ipfshash assumptions


ethers:
TokenRegistry assumes `bytes32` for ipfshash
https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/TokenRegistry.sol#L94

An ipfshash is a https://github.com/multiformats/multihash and the TokenRegistry might also be assuming a convention that the first 2 bytes of the ipfshash, which are not stored on-chain are 0x1220.

Assumptions and guidance should be documented, and/or possibly future-proofing by using `bytes` type instead.

abandeali1:
Good catch, it seems like a dynamic byte array is our best option given that there's a chance this may change in the future.

I wonder if we should do the same for `swarmHash` for the sake of future proofing as well.

maurelian:
Possibly useful:
https://www.reddit.com/r/ethdev/comments/6lbmhy/a_practical_guide_to_cheap_ipfs_hash_storage_in/

See the comments there as well. Myself, I have not thought deeply about the best approach.

PR #86, opened 2017-07-18
Make ipfs and swarm hashes dynamic byte arrays in tokenRegistry

abandeali1:
- Changes ipfsHash and swarmHash from bytes32 to dynamic byte arrays
- Removes url from token metadata (will be included in ipfs/swarm objects)
- Addresses issue #75


util/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const constants = {
     NULL_ADDRESS: '0x0000000000000000000000000000000000000000',
     NULL_HASH: '0x0000000000000000000000000000000000000000000000000000000000000000',
```
fabioberger: Is `NULL_HASH` still used anywhere? If not, we should remove it.

PR #96, opened 2017-07-21
Scalability Improvements to Custom MultiSig contract

### LINKED ISSUE #94: Scalability redesign in the MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress contract
This contract does several things not as efficiently as expected. 

Some of which are almost an anti-pattern like is the case in https://github.com/0xProject/contracts/blob/master/contracts/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol#L12 which is `assert`ing a custom function which itself `assert`s a byte-by-byte comparison of a tightly packed `bytes` array vs. a `bytes4` type signature.

Will submit a PR with a recommendation for a much cheaper and scalable design.

GNSPS:
Fixes #94.

Basic rationale is:

the bytes4FromBytes function was built act as a type cast from bytes to bytes4 using the arrays' first 4 bytes while making sure it has sufficient length.
the assert statement in the modifier is now much more verbose and easy to understand
gas costs are reduced by 2/3 approximately

(copy of #95)


GNSPS:
Forgot to mention this breaks 2 tests using the eliminated function. 

New tests would have to be created which I believe will be more streamlined, though.

(Basically just need to test the output of the first 4 bytes functions, the rest is relayed back to Solidity, so more secure!!! 😄 )

abandeali1:
Thank you! Could you please change the base branch of this PR to `frozenUpdatedCopy`? I will add tests and then make a new PR to `frozenUpdated`.

GNSPS:
Done!

Issue #100, opened 2017-07-25
Proxy contract used by TokenDistribution


ethers:
It appears that code like `exchange.PROXY_CONTRACT()` could replace https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/TokenDistributionWithRegistry.sol#L218 and the need for the TokenDistribution contract to have its own variable (which could be a mismatch against the exchange's).

It would also simplify deployment.  A deployment document would also help identify and clarify these issues.

PR #108, opened 2017-07-26
Update token sale

abandeali1:
- Make naming more explicit
- Add check to make sure order feeRecipient is null
- `initializeSale` now sets a `baseEthCapPerAddress` and the timestamp where the order becomes fillable
- Remove `setCapPerAddress`. The cap now increases automatically every 24 hours, with the increase being equal to double the previous increase (1, 3, 7, 15...)
- Fixes issue #100

### DIFF
--- a/contracts/TokenSaleWithRegistry.sol
+++ b/contracts/TokenSaleWithRegistry.sol
@@ -54,7 +54,7 @@ contract TokenSaleWithRegistry is Ownable, SafeMath {
     }
 
     modifier saleStarted() {
-        assert(startTimeInSec != 0 && block.timestamp >= startTimeInSec);
+        assert(isSaleInitialized && block.timestamp >= startTimeInSec);
         _;
     }
 
@@ -221,12 +221,14 @@ contract TokenSaleWithRegistry is Ownable, SafeMath {
     }
 
     /// @dev Calculates the ETH cap per address. The cap increases by double the previous increase at each next period. E.g 1, 3, 7, 15
-    /// @return THe current ETH cap per address.
+    /// @return The current ETH cap per address.
     function getEthCapPerAddress()
         public
         constant
-        returns (uint ethCapPerAddress)
+        returns (uint)
     {
+        if (block.timestamp < startTimeInSec || startTimeInSec == 0) return 0;
+
         uint timeSinceStartInSec = safeSub(block.timestamp, startTimeInSec);
         uint currentPeriod = safeAdd(                           // currentPeriod begins at 1
               safeDiv(timeSinceStartInSec, TIME_PERIOD_IN_SEC), // rounds down
@@ -237,7 +239,7 @@ contract TokenSaleWithRegistry is Ownable, SafeMath {
             baseEthCapPerAddress,
             safeSub(
                 2 ** currentPeriod,
-                currentPeriod
+                1
             )
         );
     }
--- a/contracts/TokenSaleWithRegistry.sol
+++ b/contracts/TokenSaleWithRegistry.sol
@@ -13,7 +13,6 @@ contract TokenSaleWithRegistry is Ownable, SafeMath {
 
     uint public constant TIME_PERIOD_IN_SEC = 1 days;
 
-    address public PROXY_CONTRACT;
     address public EXCHANGE_CONTRACT;
     address public PROTOCOL_TOKEN_CONTRACT;
     address public ETH_TOKEN_CONTRACT;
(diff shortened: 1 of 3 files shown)


test/ts/token_sale_with_registry.ts
```diff
@@ -349,51 +450,92 @@ contract('TokenSaleWithRegistry', (accounts: string[]) => {
...
+        baseEthCapPerAddress,
         { from: owner },
       );
+
+      await rpc.increaseTimeAsync(timePeriodInSec + secondsToAdd);
+      await rpc.mineBlockAsync();
+      period += 1;
+
+      const ethCapPerAddress2 = await tokenSaleWithRegistry.getEthCapPerAddress.call();
+      const expectedEthCapPerAddress2 = getEthCapPerAddress(baseEthCapPerAddress, period);
```
fabioberger: What if this calculation is equally flawed as the calculation in the smart contract? Why not hard-code the expected `ethCapPerAddress` in this test?

test/ts/token_sale_with_registry.ts
```diff
@@ -527,28 +725,127 @@ contract('TokenSaleWithRegistry', (accounts: string[]) => {
...
+        totalFilledEthValue = totalEthValue;
+
+        assert.equal(finalBalances[maker][validOrder.params.makerToken],
+                     sub(initBalances[maker][validOrder.params.makerToken], totalFilledZrxValue));
+        assert.equal(finalBalances[maker][validOrder.params.takerToken],
+                     add(initBalances[maker][validOrder.params.takerToken], totalFilledEthValue));
+        assert.equal(finalBalances[taker][validOrder.params.makerToken],
+                     add(initBalances[taker][validOrder.params.makerToken], totalFilledZrxValue));
+        assert.equal(finalTakerEthBalance, sub(sub(initTakerEthBalance, totalFilledEthValue), totalEthSpentOnGas));
       });
```
fabioberger: Let's add a test showing the any amount of ETH sent above the cap gets refunded to the sender.

contracts/TokenSaleWithRegistry.sol
```diff
@@ -144,6 +142,7 @@ contract TokenSaleWithRegistry is Ownable, SafeMath {
         require(order.taker == address(this));
         require(order.makerToken == PROTOCOL_TOKEN_CONTRACT);
         require(order.takerToken == ETH_TOKEN_CONTRACT);
+        require(order.feeRecipient == address(0));
```
abandeali1: Any reason for this?

test/ts/token_sale_with_registry.ts
abandeali1: This is implicitly tested for already in the tests that send ETH > `baseEthCapPerAddress`.

test/ts/token_sale_with_registry.ts
abandeali1: Actually, I think it probably makes more sense to hard code the `baseEthCapPerAddress` multipliers into `getEthCapPerAddress` util function.

contracts/TokenSaleWithRegistry.sol
willwarren89: `makerFee = takerFee = 0` is a necessary requirement for the order to be filled, so I think it is just a matter of making the requirement explicit.

abandeali1: If `feeRecipient = 0`, we can make fees whatever we want and it won't matter :)

test/ts/token_sale_with_registry.ts
```diff
@@ -527,28 +737,127 @@ contract('TokenSaleWithRegistry', (accounts: string[]) => {
...
+        totalFilledZrxValue = totalEthValue;
+        totalFilledEthValue = totalEthValue;
+
+        assert.equal(finalBalances[maker][validOrder.params.makerToken],
+                     sub(initBalances[maker][validOrder.params.makerToken], totalFilledZrxValue));
+        assert.equal(finalBalances[maker][validOrder.params.takerToken],
+                     add(initBalances[maker][validOrder.params.takerToken], totalFilledEthValue));
+        assert.equal(finalBalances[taker][validOrder.params.makerToken],
+                     add(initBalances[taker][validOrder.params.makerToken], totalFilledZrxValue));
+        assert.equal(finalTakerEthBalance, sub(sub(initTakerEthBalance, totalFilledEthValue), totalEthSpentOnGas));
```
fabioberger: It is verbose but eliminates the need for the reader to re-perform the calculation while reading:

```
const remainingTakerBalanceAfterFillAndGas = sub(sub(initTakerEthBalance, totalFilledEthValue), totalEthSpentOnGas);
assert.equal(finalTakerEthBalance, remainingTakerBalanceAfterFillAndGas);
```

PR #110, opened 2017-07-26
Update ZRX

abandeali1:
- Add `allowedUnlimited` mapping that maps owner => spender => bool
- Add `approveUnlimited` and `unlimitedAllowance` functions. This allows a token owner to provide an address with an unlimited allowance (ONLY DO THIS FOR CONTRACTS THAT YOU TRUST 100%). When tokens are frequently being transferred via `transferFrom` (as they are in the 0x contracts), this has significant gas savings in the long run, since the `allowed` mapping does not need to be updated for each transfer (~5k per transfer).
- `transferFrom` first checks the `allowedUnlimited` mapping, then falls back to `allowed`.
- Adds an `UnlimitedApproval` event
- Adds tests for new functionality and standard token basic functionality

### DIFF
--- a/contracts/tokens/ZRXToken.sol
+++ b/contracts/tokens/ZRXToken.sol
@@ -3,14 +3,55 @@ pragma solidity ^0.4.11;
 import "./../base/StandardToken.sol";
 
 contract ZRXToken is StandardToken {
-  
+
     uint8 constant public decimals = 18;
+    uint public totalSupply = 10**27; // 1 billion tokens, 18 decimal places
     string constant public name = "0x Protocol Token";
     string constant public symbol = "ZRX";
 
+    mapping(address => mapping(address => bool)) allowedUnlimited;
+
+    event UnlimitedApproval(address indexed _owner, address indexed _spender, bool _approval);
+
     function ZRXToken() {
-        totalSupply = 10**27; // 1 billion tokens, 18 decimal places
         balances[msg.sender] = totalSupply;
     }
 
+    function approveUnlimited(address _spender, bool _approval)
+        public
+        returns (bool)
+    {
+        allowedUnlimited[msg.sender][_spender] = _approval;
+        UnlimitedApproval(msg.sender, _spender, _approval);
+        return true;
+    }
+
+    function transferFrom(address _from, address _to, uint _value)
+        public
+        returns (bool)
+    {
+        bool isAllowedUnlimited = allowedUnlimited[_from][msg.sender];
+        if (   balances[_from] >= _value
+            && (isAllowedUnlimited || allowed[_from][msg.sender] >= _value)
+            && balances[_to] + _value >= balances[_to]
+        ) {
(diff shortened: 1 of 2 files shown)


test/ts/zrxToken.ts
```diff
@@ -0,0 +1,173 @@
...
+      const totalSupply = await zrx.totalSupply.call();
+      assert.equal(totalSupply.toString(), ownerBalance.toString());
+    });
+  });
+
+  describe('transfer', () => {
+    it('should transfer balance from sender to receiver', async () => {
+      const initOwnerBalance = await zrx.balanceOf(owner);
+      const amountToTransfer = new BigNumber(1);
+      await zrx.transfer(spender, amountToTransfer, { from: owner });
```
willwarren89: Perhaps rename `spender` to `receiver` for this test.

test/ts/zrxToken.ts
```diff
@@ -0,0 +1,173 @@
...
+      await zrx.approveUnlimited(spender, true, { from: owner });
+      await zrx.approve(spender, initSpenderAllowance, { from: owner });
+      await zrx.transferFrom(owner, spender, amountToTransfer, { from: spender });
+
+      const newSpenderAllowance = await zrx.allowance(owner, spender);
+      assert.equal(initSpenderAllowance.toString(), newSpenderAllowance.toString());
+    });
+
+    it('should transfer the correct balances if spender is not allowed unlimited but has sufficient allowance', async () => {
+      const initOwnerBalance = await zrx.balanceOf(owner);
```
fabioberger: Add assertion that spender is not allowed unlimited.

test/ts/zrxToken.ts
```diff
@@ -0,0 +1,173 @@
...
+      await zrx.transferFrom(owner, spender, amountToTransfer, { from: spender });
+
+      const newOwnerBalance = await zrx.balanceOf(owner);
+      const newSpenderBalance = await zrx.balanceOf(spender);
+
+      assert.equal(newOwnerBalance.toString(), '0');
+      assert.equal(newSpenderBalance.toString(), initOwnerBalance.toString());
+    });
+
+    it('should modify allowance if spender is not allowed unlimited but has sufficient allowance', async () => {
```
fabioberger: This test can be merged with the previous one by simply adding an assertion of the modified allowance.

contracts/tokens/ZRXToken.sol
```diff
@@ -3,14 +3,55 @@ pragma solidity ^0.4.11;
...
+                allowed[_from][msg.sender] -= _value;
+            }
+            Transfer(_from, _to, _value);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    function unlimitedAllowance(address _owner, address _spender)
```
abandeali1: I'm trying to keep the naming in line with the current ERC20 standard, just in case something like this is ever adopted.

Issue #101, opened 2017-07-25
require token amounts to be greater than 0


ethers:
Adding `require(fillTakerTokenAmount > 0);` to
https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L126

would improve safety as well as avoid spurious ERROR_ORDER_FULLY_FILLED_OR_CANCELLED  logs (when the token amount passed in is 0)
https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L142

Suggestion to do the same for `cancelOrder` https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L217

ethers:
@abandeali1 The commit you made is helpful to prevent orders with 0 token amounts, but similar checks still need to be done for `fillTakerTokenAmount` (and cancel) as described in this issue?

Suggestion: add a test where `fillTakerTokenAmount` is 0 and asserts that no logs are generated?  The test should fail since a spurious log is generated. Then adding `require(fillTakerTokenAmount > 0);` should make the test pass.

ethers:
@abandeali1 This one still needs fixing https://github.com/0xProject/contracts/blob/e51d4dcb4c8e0d93815e9d2a5c511d60ce017870/contracts/Exchange.sol#L223

Issue #105, opened 2017-07-26
Usage of `enum` is advised over 8 bit constants for error codes


GNSPS:
On
https://github.com/0xProject/contracts/blob/master/contracts/Exchange.sol#L29 and onwards you declare error Codes as 8 bit constants to address those with textual definitions throughout the contract's code, however, it is advisable you do that with an `enum` type as per Solidity specs.

This will not only serve the original purpose of this abstracted type which is more abstracted by the compiler but also save storage space and gas upon functions that use these state variables.

PR #111, opened 2017-07-26
Enum err codes and check for 0 maker/takerTokenAmounts

abandeali1:
- Closes issue #101 and issue #105

### DIFF
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -351,14 +351,14 @@ contract Exchange is SafeMath {
         returns (bool success)
     {
         for (uint i = 0; i < orderAddresses.length; i++) {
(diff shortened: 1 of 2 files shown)


contracts/Exchange.sol
```diff
@@ -349,14 +352,14 @@ contract Exchange is SafeMath {
         returns (bool success)
     {
         for (uint i = 0; i < orderAddresses.length; i++) {
-            assert(fillOrKillOrder(
```
fabioberger: Why was this assert removed?

abandeali1: It's redundant (`fillOrKillOrder` already uses assert).

Issue #107, opened 2017-07-26
`isTransferable()` function should be removed


GNSPS:
The `isTransferable()` function, referenced in:
https://github.com/0xProject/contracts/blob/master/contracts/Exchange.sol#L151
should be removed.

It does not only represent a repetition of checks against a benign (or malicious) ERC20 token and hence wasting gas (*claim 1*) but also an unnecessarily added reentrancy point (*claim 2*).

About *claim 1*:
* The function checks assume the contract's state stays intact throughout multiple calls which, by ERC20 specs, should, but in a malicious token shouldn't. (https://github.com/0xProject/contracts/blob/master/contracts/Exchange.sol#L556)
* Referring back to the point above, if the assumption about the token contract being benign is already made why check the balance again in `isTransferable()` when the same check is already made inside the ERC20 token contract? (gas is already wasted for that!)

About *claim 2*:
I couldn't, to the extent of my analyzing and testing skills, find any reentrancy *bug* but if there's any lesson *theDao* has taught us is that **accessing/changing state variables *after* an external call  in a function is not only dangerous but should be avoided at all costs**.

Such is the case on https://github.com/0xProject/contracts/blob/master/contracts/Exchange.sol#L151 as, following the external calls in `isTransferable()`, `filled[order.orderHash]` is accessed in https://github.com/0xProject/contracts/blob/master/contracts/Exchange.sol#L159.

While it doesn't represent a *bug* in itself it is a poor design choice. Any specific reason to have this check before accessing the mapping?

**Recommendation**:

If the only purpose of this check is, as seems, firing that specific event then I would propose firing that same event after checking the return boolean from the successive `transferViaProxy()` calls (https://github.com/0xProject/contracts/blob/master/contracts/Exchange.sol#L166) which represent the only call needed to check the conditions needed to fill the order.

I can only imagine the suppression of this function in such a widely used function as `fillOrder()` would represent huge gas savings.

This would also entail ditching the use of the `assert` construct which is itself wasting gas (after the `REVERT` opcode is indeed activated) and favor an `if() { Event(); revert(); }` pattern like is now present in the original check.

Or in an alternative solution have the contract call the `isTransferable()` function with a low-level Solidity `call()` function so as to not have subsequent `fillOrder()` calls interrupting a batched number of orders with bubbling up `throw`s.

abandeali1:
[...]

PR #112, opened 2017-07-26
Add gas limit to `getBalance` and `getAllowance`

abandeali1:
- Only forwards 4999 gas to `token.balanceOf` and `token.allowance` in order to prevent state changes when calling malicious tokens
- Addresses issue #107


contracts/Exchange.sol
```diff
@@ -31,6 +31,7 @@ contract Exchange is SafeMath {
     uint8 constant ERROR_ROUNDING_ERROR_TOO_LARGE = 2;          // Rounding error too large
     uint8 constant ERROR_INSUFFICIENT_BALANCE_OR_ALLOWANCE = 3; // Insufficient balance or allowance for token transfer
 
+    uint16 constant CONSTANT_EXTERNAL_CALL_GAS_LIMIT = 4999;    // Gas limit at which state update cannot occur
```
willwarren89: Perhaps rename to EXTERNAL_QUERY_GAS_LIMIT.

test/ts/exchange/core.ts
```diff
@@ -517,6 +518,22 @@ contract('Exchange', (accounts: string[]) => {
       assert.deepEqual(newBalances, balances);
     });
 
+    it('should throw if getBalance or getAllowance attempts to change state and shouldThrowOnInsufficientBalanceOrAllowance = false', async () => {
+      const maliciousToken = await MaliciousToken.new();
+      await maliciousToken.approve(Proxy.address, INIT_ALLOW, { from: taker });
```
fabioberger: Rename `INIT_ALLOW ` to `INITIAL_ALLOWANCE`

contracts/tokens/MaliciousToken.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity 0.4.11;
+
+import "./../base/StandardToken.sol";
+
+contract MaliciousToken is StandardToken {
```
ethers: not sure about Truffle but could this contract be put in a directory like `test/contracts/MaliciousToken.sol`?  (less preferable is `contracts/test/MaliciousToken.sol`)

contracts/Exchange.sol
maurelian: I'm slightly in favor of `EXTERNAL_QUERY_GAS_LIMIT`, as it seems more indicative of the reason for it's existence.

As a comment, I'd find `//  changes to state require at least 5000 gas` slightly more direct.

abandeali1: Agreed, I am generally terrible at naming variables.

contracts/tokens/MaliciousToken.sol
abandeali1: `test/contracts` didn't work, but `contracts/test` is still better than the current structure.

PR #114, opened 2017-07-27
MultiSig bytes4FromBytes

abandeali1:
- Casts the data byte array into bytes4 for comparison with the `removeAuthorizedAddress` function signature, rather than iterating through the first 4 bytes of the data array and comparing byte-by-byte.
- Fixes issue #94 
- Credit to @GNSPS, PR #96 will be closed after this is merged

### DIFF
--- a/contracts/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
+++ b/contracts/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
@@ -9,7 +9,7 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
     modifier validRemoveAuthorizedAddressTx(uint transactionId) {
         Transaction storage tx = transactions[transactionId];
         assert(tx.destination == PROXY_CONTRACT);
-        assert(isFunctionRemoveAuthorizedAddress(tx.data));
+        assert(bytes4FromBytes(tx.data) == bytes4(sha3("removeAuthorizedAddress(address)")));
         _;
     }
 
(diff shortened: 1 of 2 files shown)


contracts/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -9,7 +9,7 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
     modifier validRemoveAuthorizedAddressTx(uint transactionId) {
         Transaction storage tx = transactions[transactionId];
         assert(tx.destination == PROXY_CONTRACT);
-        assert(isFunctionRemoveAuthorizedAddress(tx.data));
+        assert(bytes4FromBytes(tx.data) == bytes4(sha3("removeAuthorizedAddress(address)")));
```
fabioberger: Let's make this a bit more readable:

```
const isFunctionRemoveAuthorizedAddress = bytes4FromBytes(tx.data) == bytes4(sha3("removeAuthorizedAddress(address)"));
assert(isFunctionRemoveAuthorizedAddress);
```

contracts/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -47,18 +47,23 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
...
     {
-        bytes4 removeAuthorizedAddressSignature = bytes4(sha3("removeAuthorizedAddress(address)"));
-        for (uint i = 0; i < 4; i++) {
-            assert(data[i] == removeAuthorizedAddressSignature[i]);
+        require(data.length >= 4);
+
+        bytes4 first4Bytes;
+
+        assembly {
+            first4Bytes := mul(div(mload(add(data, 0x20)), 0x100000000000000000000000000000000000000000000000000000000), 0x100000000000000000000000000000000000000000000000000000000)
```
fabioberger: Is there a way to break this up into multiple lines? It's currently difficult to read.

test/ts/multi_sig_with_time_lock_except_remove_auth_addr.ts
```diff
@@ -33,21 +33,14 @@ contract('MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress', (accounts: s
...
-      }
-    });
+      const first4Bytes = await multiSig.bytes4FromBytes(data);
 
-    it('should return true if data is for removeAuthorizedAddress', async () => {
-      const data = multiSigWrapper.encodeFnArgs('removeAuthorizedAddress', PROXY_ABI, [owners[0]]);
-      const isFunctionRemoveAuthorizedAddress = await multiSig.isFunctionRemoveAuthorizedAddress.call(data);
-      assert.equal(isFunctionRemoveAuthorizedAddress, true);
+      const expectedFirst4Bytes = data.slice(0, 10);
+      assert.equal(first4Bytes.length, 10);
```
fabioberger: Why is the expected length 10?

contracts/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
fabioberger: Also these constants such as `0x20` do not mean anything to me. Can we assign them to descriptive variables?

test/ts/multi_sig_with_time_lock_except_remove_auth_addr.ts
fabioberger: Also, can we move it into a well-named variable instead of repeating the constant.

abandeali1: Length of 0x + 2 hex chars per byte = 10

abandeali1:
After internal discussions, we have decided to go for code clarity over efficiency in this case. `isFunctionRemoveAuthorizedAddress` will most likely be only called once in the contract's lifetime, so efficiency isn't particularly important.

Issue #113, opened 2017-07-26
Versioning of Exchange contract


ethers:
It's not clear how an upgraded exchange contract will differentiate from this contract.

Recommendations

A version string in Exchange.sol may be useful.

Issue #104, opened 2017-07-25
Avoid "named" returns (in long functions)


ethers:
For example, replacing the named return [`returns(uint filledTakerTokenAmount)`](https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L109) with `returns (uint)` and explicitly declaring https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L140
with `uint filledTakerTokenAmount`... would make the code more clear, explicit, and less prone to bugs getting introduced (for example, if return was used instead of return 0 in that function). (The gas saving is also quite trivial.)

Some other info on "named" returns: https://github.com/ethereum/solidity/issues/1401

They are likely safer if a function is very simple like https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L572 but arguably its documentation benefit is moot.
(There's more to "named" returns such as https://github.com/ethereum/solidity/issues/1141 and we'd want code safety first.)

abandeali1:
The example you provided is unfortunately the one place we can't remove the named return, due to running against the local variable limit (the named return take up one less slot in this case).

Issue #103, opened 2017-07-25
Proper usage of `require` vs `assert`


maurelian:
After the upcoming Metropolis fork, `require` will compile to a new `REVERT` opcode, which will refund remaining gas. 

According to the [designers of solidity](https://github.com/ethereum/solidity/issues/2586#issuecomment-317841126), you should guard the code with `require` for input parameters and return values of critical calls.

Whereas an `assert` guard should be used in such a way that if it is triggered, there is a bug in the contract. 

This is inline with (a careful reading) of "[Error handling: Assert, Require, Revert and Exceptions](https://solidity.readthedocs.io/en/develop/control-structures.html#error-handling-assert-require-revert-and-exceptions)" in the docs.

Recommendation: All uses of `assert` should be reviewed according to the documentation. An example of where`require` should be used when checking return value, and normal code paths is 

https://github.com/0xProject/contracts/blob/master/contracts/Exchange.sol#L160

Issue #102, opened 2017-07-25
Lock pragmas to exact compiler version


ethers:
Example:
https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L18

Use `pragma solidity 0.4.11;` (no ^) to prevent accidental compilation with newer compilers (which 0x hasn't been tested against).

PR #117, opened 2017-07-27
Small fixes

abandeali1:
- Make compiler version static (issue #102)
- Change improper uses of `assert` to `require` (issue #103)
- Remove (most) named returns (issue #104)
- Add version string to Exchange (issue #113)

### DIFF
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -33,7 +33,8 @@ contract Exchange is SafeMath {
         INSUFFICIENT_BALANCE_OR_ALLOWANCE // Insufficient balance or allowance for token transfer
     }
 
-    uint16 constant EXTERNAL_QUERY_GAS_LIMIT = 4999;    // Changes to state require at least 5000 gas
+    string constant public EXCHANGE_VERSION = "1.0.0";
+    uint16 constant public EXTERNAL_QUERY_GAS_LIMIT = 4999;    // Changes to state require at least 5000 gas
 
     address public ZRX_TOKEN_CONTRACT;
     address public PROXY_CONTRACT;
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -15,7 +15,7 @@
   limitations under the License.
 
 */
-pragma solidity ^0.4.11;
+pragma solidity 0.4.11;
 
 import "./Proxy.sol";
 import "./base/Token.sol";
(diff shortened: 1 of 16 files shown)


contracts/Exchange.sol
```diff
@@ -33,7 +33,8 @@ contract Exchange is SafeMath {
         INSUFFICIENT_BALANCE_OR_ALLOWANCE // Insufficient balance or allowance for token transfer
     }
 
-    uint16 constant EXTERNAL_QUERY_GAS_LIMIT = 4999;    // Changes to state require at least 5000 gas
+    string constant public EXCHANGE_VERSION = "1.0.0";
```
fabioberger: I think we can name this simple `VERSION` so a call to it would look like:

```
const exchangeVersion = await exchange.VERSION.call();
```

contracts/TokenRegistry.sol
```diff
@@ -219,12 +219,12 @@ contract TokenRegistry is Ownable {
...
-            string symbol,
-            uint8 decimals,
-            bytes ipfsHash,
-            bytes swarmHash
+            address,
+            string,
+            string,
+            uint8,
+            bytes,
+            bytes
```
fabioberger: Let's add the variable names next to each of these as comments. i.e:

```
address, // tokenAddress
string, //name
...
```

contracts/TokenRegistry.sol
```diff
@@ -245,12 +245,12 @@ contract TokenRegistry is Ownable {
...
```
fabioberger: Ditto adding value names as comments next to each since this is still helpful information that makes this more readable.

contracts/TokenRegistry.sol
```diff
@@ -264,12 +264,12 @@ contract TokenRegistry is Ownable {
...
```
fabioberger: Ditto above

contracts/base/MultiSigWallet.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.8;
+pragma solidity 0.4.11;
```
fabioberger: Let's check if a newer version of this contract exists with the newer version of Solidity. We should also set up a system for tracking bug fixes in any third-party smart contracts that we are using.

abandeali1: Hmm, looks like the actual contract uses 0.4.4, we should probably stick with that.

Issue #98, opened 2017-07-22
isRoundingError is not limited to 0.1%


ethers:
`isRoundingError` https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L468 states "Checks if rounding error > 0.1%."

Exchange.sol has https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol
```
        if (isRoundingError(filledTakerTokenAmount, order.takerTokenAmount, order.makerTokenAmount)) {
            LogError(ERROR_ROUNDING_ERROR_TOO_LARGE, order.orderHash);
            return 0;
        }
```
Presumably so that the taker's maximum loss is 0.1%.

But if a maker is offering 1000 tokens for 7 of taker's tokens, and a taker submits 1 taker token, then the taker's loss is larger than 0.001... Taker should have received 142.857 tokens and received only 142 tokens...

Fixes to this should include tests: Related #92

willwarren89:
To determine the amount of `makerToken` that `taker` will receive in the case of a partial fill, we must solve for `x = a/b * y` where

- y = the total amount of makerToken offered by Maker for a given order
- b = the total amount of takerToken requested by Maker for a given order
- a = the amount of takerToken that Taker is attempting to sell/fill (`0 <= a/b <= 1`) 
- all values are of type `uint256`

**It follows that**
- integer division will always produce a truncation error when `(ay mod b) != 0`
- absolute truncation is given by remainder `R = (ay mod b)`
- normalized truncation error (as % of partial fill amount) is given by `e = R/x = b (ay mod b)/ay`

**Analysis**
...

ethers:
@willwarren89 Can you please clarify prior comment?  Is your goal `R/x` to be <= 0.001, where `x` is the `floor(a*y / b)` ?  That's my understanding but this doesn't seem to match https://github.com/0xProject/contracts/pull/129/files#diff-e1e51de7476e0cab7ae50ddcbd4c0ff1R105

Is it possible to step back and also focus on the "customer scenario" and give examples of orders that should not go through because of error too large?

Could those tests be added?  More specifically, can the following "boundary" scenarios/tests be added:
```
error of 9/10000, order should not succeed
error of 10/10000, order should succeed
error of 11/10000, order should succeed
```

[...]

Issue #92, opened 2017-07-20
Missing tests for isRoundingError and getPartialAmount


ethers:
Ideally `isRoundingError` and `getPartialAmount` https://github.com/0xProject/contracts/blob/888d5a02573572240f4c55e03238be603c13c469/contracts/Exchange.sol#L473-L490 should have dedicated unit tests (rather than indirectly tested by the fill functions, because they can be called by other code).  Test the "limits" of each function and document when the behavior changes.

[`mulmod` should be able handle much larger values (http://solidity.readthedocs.io/en/develop/units-and-global-variables.html#mathematical-and-cryptographic-functions) than `getPartialAmount` which will fail an assert at some point.]

ethers:
Much larger values should be tested to determine the limits of each function and when the behavior deviates from what's expected.

PR #129, opened 2017-07-28
Fix isRoundingError

abandeali1:
- Closes issues #98 and #92

### DIFF
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -490,7 +490,20 @@ contract Exchange is SafeMath {
         constant
         returns (bool)
     {
-        return (target < 10**3 && mulmod(target, numerator, denominator) != 0);
+        if (mulmod(target, numerator, denominator) == 0) return false; // No rounding error.
+        uint partialAmountWithErr = safeMul(getPartialAmount(numerator, denominator, target), 1000000);
+        uint partialAmountWithoutErr = safeDiv(
+            safeMul(
+                safeMul(numerator, target),
+                1000000
+            ),
+            denominator
+        );
+        uint errPercentageTimes1000 = safeDiv(
+            safeSub(partialAmountWithoutErr, partialAmountWithErr), // Absolute rounding error, times 1,000,000
+            safeDiv(partialAmountWithoutErr, 1000)                  // Amount being filled, times 1,000
+        );
+        return errPercentageTimes1000 > 1;
     }
 
     /// @dev Calculates partial value given a numerator and denominator.
--- a/contracts/Exchange.sol
+++ b/contracts/Exchange.sol
@@ -491,7 +491,9 @@ contract Exchange is SafeMath {
         returns (bool)
     {
         if (mulmod(target, numerator, denominator) == 0) return false; // No rounding error.
+        // (numerator * target / denumerator) * 1000000
         uint partialAmountWithErr = safeMul(getPartialAmount(numerator, denominator, target), 1000000);
+        // (numerator * target * 1000000 / denumerator)
         uint partialAmountWithoutErr = safeDiv(
             safeMul(
                 safeMul(numerator, target),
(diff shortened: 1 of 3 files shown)


test/ts/exchange/helpers.ts
```diff
@@ -76,4 +77,56 @@ contract('Exchange', (accounts: string[]) => {
...
+    it('should round down', async () => {
+      const numerator = new BigNumber(2);
+      const denominator = new BigNumber(3);
+      const target = new BigNumber(10);
+
+      const partialAmount = await exWrapper.getPartialAmountAsync(numerator, denominator, target);
+      const expectedPartialAmount = '6';
+      assert.equal(partialAmount.toString(), expectedPartialAmount);
+    });
+  });
```
LogvinovLeon: I'll add one more test for rounding of 0.5. Different implementations differ and we want to make sure it's behaving like expected.
```
const numerator = new BigNumber(1);
const denominator = new BigNumber(20);
const target = new BigNumber(10);
```

test/ts/exchange/helpers.ts
```diff
@@ -76,4 +77,56 @@ contract('Exchange', (accounts: string[]) => {
...
+      assert.equal(isRoundingError, false);
+    });
+
+    it('should return false when there is rounding error <= 0.1%', async () => {
+      // randomly generated numbers
+      const numerator = new BigNumber(76564);
+      const denominator = new BigNumber(676373677);
+      const target = new BigNumber(105762562);
+      // rounding error = ((76564*105762562/676373677) - floor(76564*105762562/676373677)) / (76564*105762562/676373677) = 0.0007%
+      const isRoundingError = await exWrapper.isRoundingErrorAsync(numerator, denominator, target);
```
LogvinovLeon: I'll rename `exWrapper` to `exchangeWrapper`

contracts/Exchange.sol
```diff
@@ -490,7 +490,20 @@ contract Exchange is SafeMath {
...
+                safeMul(numerator, target),
+                1000000
+            ),
+            denominator
+        );
+        uint errPercentageTimes1000 = safeDiv(
+            safeSub(partialAmountWithoutErr, partialAmountWithErr), // Absolute rounding error, times 1,000,000
+            safeDiv(partialAmountWithoutErr, 1000)                  // Amount being filled, times 1,000
+        );
+        return errPercentageTimes1000 > 1;
```
LogvinovLeon: In contrast with the mutisig's bytes to bytes4 this function is executed on every fill and that's why gas cost matters. What's the gas cost of this one?
Can we somehow optimize it?
Maybe it's better to leave it for v2.

contracts/Exchange.sol
```diff
@@ -490,7 +490,20 @@ contract Exchange is SafeMath {
...
+            safeMul(
+                safeMul(numerator, target),
+                1000000
+            ),
+            denominator
+        );
+        uint errPercentageTimes1000 = safeDiv(
+            safeSub(partialAmountWithoutErr, partialAmountWithErr), // Absolute rounding error, times 1,000,000
+            safeDiv(partialAmountWithoutErr, 1000)                  // Amount being filled, times 1,000
+        );
```
LogvinovLeon: With all the safeDivs and safeSubs this logic is cryptic.
I'll add a comment explaining what's happening similar to the one we have in tests.

contracts/Exchange.sol
abandeali1: I ran a few tests and the difference seems to be negligible, surprisingly. Probably worth looking into a bit more.

test/ts/exchange/helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+      const isRoundingError = await exchangeWrapper.isRoundingErrorAsync(numerator, denominator, target);
+      assert.equal(isRoundingError, false);
+    });
+
+    it('should return false when there is rounding error <= 0.1%', async () => {
+      // randomly generated numbers
+      const numerator = new BigNumber(76564);
+      const denominator = new BigNumber(676373677);
+      const target = new BigNumber(105762562);
+      // rounding error = ((76564*105762562/676373677) - floor(76564*105762562/676373677)) / (76564*105762562/676373677) = 0.0007%
```
ethers: is this computation supposed to be the same as https://github.com/0xProject/contracts/issues/98 ?

PR #146, opened 2017-08-03
unfillable order test case

glambeth:
demonstrates an unfillable order

### DIFF
--- a/test/ts/exchange/core.ts
+++ b/test/ts/exchange/core.ts
@@ -115,6 +115,30 @@ contract('Exchange', (accounts: string[]) => {
       order = await orderFactory.newSignedOrderAsync();
     });
 
+     it('should create an unfillable order', async () => {
+      order = await orderFactory.newSignedOrderAsync({
+        makerTokenAmount: new BigNumber(1001),
+        takerTokenAmount: new BigNumber(3)
+      });
+
+      const filledTakerTokenAmountBefore = await exchange.filled.call(order.params.orderHashHex);
+      assert.equal(filledTakerTokenAmountBefore, 0, 'filledAmountMBefore should be 0');
+      
+      const fillTakerTokenAmount1 = new BigNumber(2);
+      await exWrapper.fillOrderAsync(order, taker, { fillTakerTokenAmount: fillTakerTokenAmount1 });
+
+      const filledTakerTokenAmountAfter1 = await exchange.filled.call(order.params.orderHashHex);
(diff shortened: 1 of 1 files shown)


test/ts/exchange/core.ts
```diff
@@ -115,6 +115,30 @@ contract('Exchange', (accounts: string[]) => {
...
+      
+      const fillTakerTokenAmount1 = new BigNumber(2);
+      await exWrapper.fillOrderAsync(order, taker, { fillTakerTokenAmount: fillTakerTokenAmount1 });
+
+      const filledTakerTokenAmountAfter1 = await exchange.filled.call(order.params.orderHashHex);
+      assert.equal(filledTakerTokenAmountAfter1, fillTakerTokenAmount1.toString(),
+                   'filledTakerTokenAmountAfter1 should be same as fillTakerTokenAmount1');
+                     
+      const fillTakerTokenAmount2 = new BigNumber(1);
+      await exWrapper.fillOrderAsync(order, taker, { fillTakerTokenAmount: fillTakerTokenAmount2 });
```
ethers: is there a way to get the return value of fillOrder?  the idea is to assert that the return value is 0

test/ts/exchange/core.ts
```diff
@@ -115,6 +115,30 @@ contract('Exchange', (accounts: string[]) => {
...
+      });
+
+      const filledTakerTokenAmountBefore = await exchange.filled.call(order.params.orderHashHex);
+      assert.equal(filledTakerTokenAmountBefore, 0, 'filledAmountMBefore should be 0');
+      
+      const fillTakerTokenAmount1 = new BigNumber(2);
+      await exWrapper.fillOrderAsync(order, taker, { fillTakerTokenAmount: fillTakerTokenAmount1 });
+
+      const filledTakerTokenAmountAfter1 = await exchange.filled.call(order.params.orderHashHex);
+      assert.equal(filledTakerTokenAmountAfter1, fillTakerTokenAmount1.toString(),
```
abandeali1: This should be `assert.equal(filledTakerTokenAmount.toString(), fillTakerTokenAmount1.toString(), ...`
Other than that, LGTM!

test/ts/exchange/core.ts
abandeali1: You can do `exchange.fillOrder.call(...)`, but it requires passing in the order parameters in their array formats. Another option would be to check if the correct error event was fired.

PR #152, opened 2017-08-09
TokenRegistry improvements

jakub-wojciechowski:
As discussed with @abandeali1 in emails: 

1. Checkers for duplicated addresses, names and symbols used in modifiers:
   require(tokens[_token].token != address(0));

check for "0x" value which is both a valid address and default empty value. As "0x" is used as a special address for avoiding fee payments it may be tempting to register it as a special "null token". Later, the name or symbol may be overwritten to point to a different malicious address that will collect all of the "0x" fee payment. It'd be a better practice to either prevent registering "0x" address or distinguishing between null and "0x" value.

2. Function removeToken(address _token) 
     is getting more and more expensive with every new token added to the storage (820gas per token). Although the hard cap is pretty high (8k with current block limit) it's correctness depends on block limit set be miners. It can easily fixed to have constant running cost.

### DIFF
--- a/test/ts/token_registry.ts
+++ b/test/ts/token_registry.ts
@@ -225,7 +225,7 @@ contract('TokenRegistry', (accounts: string[]) => {
     describe('removeToken', () => {
       it('should throw if not called by owner', async () => {
         try {
-          await tokenReg.removeToken(token1.address, { from: notOwner });
+          await tokenReg.removeToken(token1.address, 0, { from: notOwner });
           throw new Error('removeToken succeeded when it should have thrown');
         } catch (err) {
           testUtil.assertThrow(err);
@@ -233,7 +233,7 @@ contract('TokenRegistry', (accounts: string[]) => {
       });
 
       it('should remove token metadata when called by owner', async () => {
-        const res = await tokenReg.removeToken(token1.address, { from: owner });
+        const res = await tokenReg.removeToken(token1.address, 0, { from: owner });
         assert.equal(res.logs.length, 1);
         const tokenData = await tokenRegWrapper.getTokenMetaDataAsync(token1.address);
         assert.deepEqual(tokenData, nullToken);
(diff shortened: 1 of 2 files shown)


test/ts/token_registry.ts
```diff
@@ -216,28 +225,38 @@ contract('TokenRegistry', (accounts: string[]) => {
...
+          await tokenReg.removeToken(nullToken.address, 0, { from: owner });
           throw new Error('removeToken succeeded when it should have failed');
         } catch (err) {
           testUtil.assertThrow(err);
         }
       });
+
+      it('should throw if token at given index does not match address', async () => {
+        try {
+          await tokenReg.removeToken(nullToken.address, 1, { from: owner });
```
abandeali1: Can we try adding and then removing the token at the wrong index? This currently will throw an error from attempting to remove a null address, so it will never actually get the the point where it checks the token the index.

jakub-wojciechowski: Sorry, I forgot about the `tokenExists` modifier. I've pushed the fix.

test/ts/token_registry.ts
```diff
@@ -216,28 +225,40 @@ contract('TokenRegistry', (accounts: string[]) => {
     describe('removeToken', () => {
       it('should throw if not called by owner', async () => {
         try {
-          await tokenReg.removeToken(token1.address, { from: notOwner });
+          await tokenReg.removeToken(token1.address, 0, { from: notOwner });
```
fabioberger: Let's not inline the index here, it's very unreadable. Let's assign to a descriptive variable name instead.

test/ts/token_registry.ts
```diff
@@ -216,28 +225,40 @@ contract('TokenRegistry', (accounts: string[]) => {
...
           throw new Error('removeToken succeeded when it should have thrown');
         } catch (err) {
           testUtil.assertThrow(err);
         }
       });
 
       it('should remove token metadata when called by owner', async () => {
-        const res = await tokenReg.removeToken(token1.address, { from: owner });
+        const res = await tokenReg.removeToken(token1.address, 0, { from: owner });
```
fabioberger: ditto above.

test/ts/token_registry.ts
```diff
@@ -216,28 +225,40 @@ contract('TokenRegistry', (accounts: string[]) => {
...
         assert.equal(res.logs.length, 1);
         const tokenData = await tokenRegWrapper.getTokenMetaDataAsync(token1.address);
         assert.deepEqual(tokenData, nullToken);
       });
 
       it('should throw if token does not exist', async () => {
         try {
-          await tokenReg.removeToken(nullToken.address, { from: owner });
+          await tokenReg.removeToken(nullToken.address, 0, { from: owner });
```
fabioberger: ditto above.

test/ts/token_registry.ts
```diff
@@ -216,28 +225,40 @@ contract('TokenRegistry', (accounts: string[]) => {
...
         } catch (err) {
           testUtil.assertThrow(err);
         }
       });
+
+      it('should throw if token at given index does not match address', async () => {
+        await tokenRegWrapper.addTokenAsync(token2, owner);
+
+        try {
+          await tokenReg.removeToken(token2.address, 0, { from: owner });
```
fabioberger: ditto above.

abandeali1:
Made the final changes in #153, closing this PR.

PR #164, opened 2017-09-07
Compile script

abandeali1:
- Adds a script to compile the contracts and write artifacts in a new format (see `scripts/schemas/artifact_schema.ts`). The script also supports projects that use multiple Solidity versions. Adding new versions of Solidity requires adding the new solc binary in `scripts/solc/solc_bin` and updating mappings in `scripts/solc/bin_paths.ts`


package.json
```diff
@@ -7,9 +7,9 @@
     "test": "test"
   },
   "scripts": {
-    "transpile": "rm -rf ./transpiled; copyfiles ./build/**/* ./transpiled; tsc",
+    "transpile": "rm -rf ./transpiled; tsc; copyfiles ./build/**/* ./scripts/solc/solc_bin/* ./transpiled;",
```
LogvinovLeon: Why did we move the `tsc` command?

package.json
```diff
@@ -30,13 +30,15 @@
...
+    "@types/semver": "^5.4.0",
     "copyfiles": "^1.2.0",
     "dirty-chai": "^2.0.0",
-    "solc": "^0.4.11",
     "ethereumjs-testrpc": "4.0.1",
+    "solc": "^0.4.15",
     "truffle": "3.4.3",
     "typescript": "^2.2.2",
-    "web3-typescript-typings": "0.0.3"
+    "web3-typescript-typings": "0.0.3",
```
LogvinovLeon: Please update to the new version. A lot had changed since 0.0.3

scripts/schemas/artifact_schema.ts
```diff
@@ -0,0 +1,18 @@
...
+    contract_name: string;
+    networks: ContractNetworks;
+};
+
+export interface ContractNetworks {
+    [key: number]: ContractData;
+};
+
+export interface ContractData {
+    solc_version: string;
```
LogvinovLeon: Let's make it a string enum and reuse it in `BinaryPaths`

scripts/schemas/artifact_schema.ts
```diff
@@ -0,0 +1,18 @@
...
+    optimizer_runs: number;
+    keccak256: string;
+    abi: any[];
```
LogvinovLeon: `Web3. ContractAbi`

scripts/schemas/artifact_schema.ts
```diff
@@ -0,0 +1,18 @@
+export interface ContractArtifact {
```
LogvinovLeon: Those are not schemas. Those are just types. Let's put them in types file as we do in other repos and import them from there

scripts/solc/bin_paths.ts
```diff
@@ -0,0 +1,12 @@
+interface BinaryPaths {
+    [key: string]: string;
```
LogvinovLeon: Use the string enum declared above

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
+import * as fs from 'fs';
+import * as solc from 'solc';
+import * as promisify from 'es6-promisify';
```
LogvinovLeon: If you define the real type of this module (Copy from other repos) - you'll not be able to import it like that and 'll need a require import

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+import * as path from 'path';
+import * as _ from 'lodash';
+import ethUtil =  require('ethereumjs-util');
```
LogvinovLeon: But this is a module and you can import it normally

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+import {binPaths} from './solc/bin_paths';
+import {ContractArtifact, ContractNetworks, ContractData} from './schemas/artifact_schema';
+import {CompilerOptions, ContractSources} from './schemas/compiler_schema';
+
+const readdirAsync = promisify(fs.readdir);
```
LogvinovLeon: https://www.npmjs.com/package/promisify-node

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+export class Compiler {
+    private contractsDir: string;
+    private networkId: number;
+    private optimizerRuns: number;
+    private artifactsDir: string;
+
+    constructor(options: CompilerOptions) {
+        this.contractsDir = options.contractsDir;
+        this.networkId = options.networkId;
+        this.optimizerRuns = options.optimizerRuns;
```
LogvinovLeon: What about `optimizerEnabled`?

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+import * as _ from 'lodash';
+import ethUtil =  require('ethereumjs-util');
+import {binPaths} from './solc/bin_paths';
+import {ContractArtifact, ContractNetworks, ContractData} from './schemas/artifact_schema';
+import {CompilerOptions, ContractSources} from './schemas/compiler_schema';
+
+const readdirAsync = promisify(fs.readdir);
+const readFileAsync = promisify(fs.readFile);
+const writeFileAsync = promisify(fs.writeFile);
+const doesPathExist = promisify(fs.access);
```
LogvinovLeon: Use `fs.existSync` https://stackoverflow.com/a/4482701/3546986

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+        try {
+            const versionPragma = source.match(/(?:solidity\s)([^]?[0-9]{1,2}[.][0-9]{1,2}[.][0-9]{1,2})/)[0];
+            const solcVersion = versionPragma.replace('^', '').slice(9);
+            return solcVersion;
+        } catch (err) {
+            throw new Error('Could not find Solidity version in source');
+        }
+    }
+
+    public getNormalizedErrMsg(errMsg: string): string {
```
LogvinovLeon: Can we please write either a test or an example for it?
It's kinda clear, what this does, but not 100%

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+            const baseContract = path.basename(errPath);
+            const normalizedErrMsg = errMsg.replace(errPath, baseContract);
+            return normalizedErrMsg;
+        } catch (err) {
+            throw new Error('Could not find a path in error message');
+        }
+
+    }
+
+    public async createArtifactsDirIfDoesNotExist(): Promise<void> {
```
LogvinovLeon: private

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+            const contentPath = `${dirPath}/${name}`;
+            if (path.extname(name) === '.sol') {
+                try {
+                    sources[name] = await readFileAsync(contentPath, {encoding: 'utf8'});
+                    log(`Reading ${name} source...`);
+                } catch (err) {
+                    log(`Could not find file at ${contentPath}`);
+                }
+            } else {
+                const nestedSources = await this.generateSources(contentPath);
```
LogvinovLeon: Let's check if it's a directory. I can imagine - that someone will put non-sol files in the source dir.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+        this.networkId = options.networkId;
+        this.optimizerRuns = options.optimizerRuns;
+        this.artifactsDir = options.artifactsDir;
+    }
+
+    public async compileAll(): Promise<void> {
+        await this.createArtifactsDirIfDoesNotExist();
+        const sources: ContractSources = await this.generateSources(this.contractsDir);
+
+        const findImports = (importPath: string): any => {
```
LogvinovLeon: return type

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+            const contractBaseName = path.basename(importPath);
+            const source = sources[contractBaseName];
+            return {contents: source};
+        };
+
+        const contractBaseNames = _.keys(sources);
+        const warnings: any = {};
```
LogvinovLeon: type

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                const solcInstance = solc.setupMethods(solcBin);
+
+                log(`Compiling ${contractBaseName}...`);
+                const compiled = solcInstance.compile({sources: input}, this.optimizerRuns, findImports);
+
+                if (!_.isUndefined(compiled.errors)) {
+                    _.each(compiled.errors, errMsg => {
+                        const normalizedErrMsg = this.getNormalizedErrMsg(errMsg);
+                        if (_.isUndefined(warnings[normalizedErrMsg])) {
+                            warnings[normalizedErrMsg] = true;
```
LogvinovLeon: Why do we record errors as warnings?

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                    optimizer_runs: this.optimizerRuns,
+                    abi: JSON.parse(compiled.contracts[contractIdentifier].interface),
+                    unlinked_binary: `0x${compiled.contracts[contractIdentifier].bytecode}`,
+                    updated_at: Date.now(),
+                };
+
+                let newArtifact: ContractArtifact;
+                if (!_.isUndefined(currentArtifactString)) {
+                    const newNetworks: ContractNetworks = _.assign({}, oldNetworks, {[this.networkId]: contractData});
+                    newArtifact = _.assign({}, currentArtifact, {networks: newNetworks});
```
LogvinovLeon:
```
newArtifact = _.assign({}, currentArtifact, {
    networks: {
        ... oldNetworks,
        {
            [this.networkId]: contractData,
        },
    },
});
```

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                } else {
+                    newArtifact = {
+                        contract_name: contractName,
+                        networks: {[this.networkId]: contractData},
```
LogvinovLeon: Our convention is to always have every object property on a separate line

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+
+const readdirAsync = promisify(fs.readdir);
+const readFileAsync = promisify(fs.readFile);
+const writeFileAsync = promisify(fs.writeFile);
+const doesPathExist = promisify(fs.access);
+const mkdirAsync = promisify(fs.mkdir);
+
+const log = console.log;
+
+const JSON_REPLACER: any = null;
```
LogvinovLeon: The type can be `null`

scripts/index.ts
```diff
@@ -0,0 +1,45 @@
+import * as yargs from 'yargs';
```
LogvinovLeon: https://www.npmjs.com/package/@types/yargs

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                    };
+                }
+
+                await writeFileAsync(currentArtifactPath, JSON.stringify(newArtifact, JSON_REPLACER, JSON_SPACES));
+                log(`${contractBaseName} artifact saved!`);
+            }
+        });
+    }
+
+    public async generateSources(dirPath: string): Promise<ContractSources> {
```
LogvinovLeon: private

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                }
+            } else {
+                const nestedSources = await this.generateSources(contentPath);
+                sources = _.assign({}, sources, nestedSources);
+            }
+        }
+        return sources;
+    }
+
+    public parseSolidityVersion(source: string): string {
```
LogvinovLeon: private

scripts/Compiler.ts
LogvinovLeon: private

LogvinovLeon:
Those are a lot of comments, but most of them if not all are nitpicks.
I'm very excited about this change!

fabioberger:
@abandeali1 can we add `The script also supports projects that use multiple Solidity versions. Adding new versions of Solidity requires adding the new solc binary in scripts/solc/solc_bin and updating mappings in scripts/solc/bin_paths.ts` to a Quip doc on how to use this tool?

contracts/Migrations.sol
```diff
@@ -1,23 +0,0 @@
...
-
-    function setCompleted(uint completed) restricted {
-        last_completed_migration = completed;
-    }
-
-    function upgrade(address new_address) restricted {
-        Migrations upgraded = Migrations(new_address);
-        upgraded.setCompleted(last_completed_migration);
-    }
-}
```
fabioberger: 💯

globals.d.ts
```diff
@@ -1,6 +1,8 @@
 declare module 'bn.js';
 declare module 'ethereumjs-abi';
 declare module 'es6-promisify';
+declare module 'yargs';
+declare module 'solc';
```
fabioberger: We should start being better about adding type defs for modules that don't have them. Take a look at 0x.js `globals.d.ts`, we hardly have any `declare module`'s at the top of the file.

scripts/Compiler.ts
fabioberger: can you explain what your comment means @LogvinovLeon, we aren't mind readers here 📦

scripts/Compiler.ts
fabioberger: We shouldn't be using `null` anywhere in our codebase. Why is this set to `null` instead of undefined?

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+const readdirAsync = promisify(fs.readdir);
+const readFileAsync = promisify(fs.readFile);
+const writeFileAsync = promisify(fs.writeFile);
+const doesPathExist = promisify(fs.access);
+const mkdirAsync = promisify(fs.mkdir);
+
+const log = console.log;
+
+const JSON_REPLACER: any = null;
+const JSON_SPACES = 4;
```
fabioberger: I'd prefer `NUMBER_OF_JSON_SPACES`

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+import {ContractArtifact, ContractNetworks, ContractData} from './schemas/artifact_schema';
+import {CompilerOptions, ContractSources} from './schemas/compiler_schema';
+
+const readdirAsync = promisify(fs.readdir);
+const readFileAsync = promisify(fs.readFile);
+const writeFileAsync = promisify(fs.writeFile);
+const doesPathExist = promisify(fs.access);
+const mkdirAsync = promisify(fs.mkdir);
+
+const log = console.log;
```
fabioberger: Um... don't we have a `utils.consoleLog` utility that you could use? I'm not a fan of this convention.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+    private optimizerRuns: number;
+    private artifactsDir: string;
+
+    constructor(options: CompilerOptions) {
+        this.contractsDir = options.contractsDir;
+        this.networkId = options.networkId;
+        this.optimizerRuns = options.optimizerRuns;
+        this.artifactsDir = options.artifactsDir;
+    }
+
```
fabioberger: No spaces between methods in a class def is our convention.

scripts/Compiler.ts
fabioberger: Why is this method declared within another method? I'd prefer we avoid this pattern and instead define it in the top-scope and pass in any args it needs.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+    public async compileAll(): Promise<void> {
+        await this.createArtifactsDirIfDoesNotExist();
+        const sources: ContractSources = await this.generateSources(this.contractsDir);
+
+        const findImports = (importPath: string): any => {
+            const contractBaseName = path.basename(importPath);
+            const source = sources[contractBaseName];
+            return {contents: source};
```
fabioberger: our convention is to always return a named variable. It makes the code more readable.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+        };
+
+        const contractBaseNames = _.keys(sources);
+        const warnings: any = {};
+
+        _.each(contractBaseNames, async contractBaseName => {
```
fabioberger: `async contractBaseName` should have parens and type def.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+            const source = sources[contractBaseName];
+            const contractName = path.basename(contractBaseName, '.sol');
```
fabioberger: Let's set `'.sol'` to a constant variable named `SOLIDITY_FILE_PREFIX`

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                currentArtifactString = await readFileAsync(currentArtifactPath, {encoding: 'utf8'});
+                currentArtifact = JSON.parse(currentArtifactString);
+                oldNetworks = currentArtifact.networks;
+                const oldNetwork: ContractData = oldNetworks[this.networkId];
+                if (!_.isUndefined(oldNetwork) && oldNetwork.keccak256 === sourceHash &&
+                    oldNetwork.optimizer_runs === this.optimizerRuns) {
+                    shouldCompile = false;
+                } else {
+                    shouldCompile = true;
+                }
```
fabioberger: Can re-write as:

```
shouldCompile = _.isUndefined(oldNetwork) ||
                             oldNetwork.keccak256 !== sourceHash ||
                             oldNetwork.optimizer_runs !== this.optimizerRuns;
```

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+            } catch (err) {
+                shouldCompile = true;
+            }
+
+            if (shouldCompile) {
+                const input = {[contractBaseName]: source};
+                const solcVersion = this.parseSolidityVersion(input[contractBaseName]);
```
fabioberger: Isn't `input[contractBaseName]` the same as `source`?

scripts/Compiler.ts
fabioberger: Yeah, our convention is no JS objects all on one line. Please use new lines as it helps with code readability.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                    const newNetworks: ContractNetworks = _.assign({}, oldNetworks, {[this.networkId]: contractData});
+                    newArtifact = _.assign({}, currentArtifact, {networks: newNetworks});
+                } else {
+                    newArtifact = {
+                        contract_name: contractName,
+                        networks: {[this.networkId]: contractData},
+                    };
+                }
+
+                await writeFileAsync(currentArtifactPath, JSON.stringify(newArtifact, JSON_REPLACER, JSON_SPACES));
```
fabioberger: Assign `JSON.stringify(newArtifact, JSON_REPLACER, JSON_SPACES)` to a variable.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+        let sources: ContractSources = {};
+        let dirContents: string[];
+        try {
+            dirContents = await readdirAsync(dirPath);
+        } catch (err) {
+            throw new Error(`No directory found at ${dirPath}`);
+        }
+        for (const name of dirContents) {
+            const contentPath = `${dirPath}/${name}`;
+            if (path.extname(name) === '.sol') {
```
fabioberger: Ah ha! We re-use `.sol` here, definitively should be a constant.

scripts/Compiler.ts
```diff
@@ -0,0 +1,176 @@
...
+                const nestedSources = await this.generateSources(contentPath);
+                sources = _.assign({}, sources, nestedSources);
+            }
+        }
+        return sources;
+    }
+
+    public parseSolidityVersion(source: string): string {
+        try {
+            const versionPragma = source.match(/(?:solidity\s)([^]?[0-9]{1,2}[.][0-9]{1,2}[.][0-9]{1,2})/)[0];
```
fabioberger: Where did you get this Regex from? If from somewhere other then your head, please add a source comment above:

```
Source: https://...
```

scripts/schemas/compiler_schema.ts
```diff
@@ -0,0 +1,10 @@
+export interface CompilerOptions {
+    contractsDir: string;
+    networkId: number;
+    optimizerRuns: number;
+    artifactsDir: string;
+}
+
+export interface ContractSources {
```
fabioberger: These are also type defs, not schemas. Please move to `types.ts`.

scripts/Compiler.ts
abandeali1: Hmm, it's annoying but the replacer as `void` doesn't match the `JSON.stringify` type definition (but `null` does).

scripts/Compiler.ts
abandeali1: No source, I created it.

scripts/Compiler.ts
abandeali1: Re not using `null`, `undefined` does not match the built in type definition of `JSON.stringify`.

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+    private artifactsDir: string;
+
+    constructor(options: CompilerOptions) {
+        this.contractsDir = options.contractsDir;
+        this.networkId = options.networkId;
+        this.optimizerEnabled = options.optimizerEnabled;
+        this.artifactsDir = options.artifactsDir;
+    }
+    /**
+     * Compiles all Solidity files found it contractsDir and writes JSON artifacts to artifactsDir.
```
fabioberger: typo: `found in`

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+     */
+    public async compileAll(): Promise<void> {
+        await this.createArtifactsDirIfDoesNotExist();
+        const sources: ContractSources = await this.getContractSources(this.contractsDir);
+        const findImports: (importPath: string) => ImportContents = this.createFindImports(sources);
+
+        const contractBaseNames = _.keys(sources);
+        const errors: SolcErrors = {};
```
fabioberger: let's rename this to `solcErrors`

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+
+        _.each(contractBaseNames, async (contractBaseName: string): Promise<void> => {
```
fabioberger: I'm pretty sure that `lodash`'s `_.each` will not `await` the `async` method it is calling on each `contractBaseName`. There is the chance this script exists before those methods finish running. What you want to use here is:

```ts
for (const contractBaseName as contractBaseNames) {
  await this.compileContractAsync(contractBaseName);
}
```

And move this anonymous async method into a private method on this class. This way you are sure that each compilation completes before the script exists. If you want to run these compilations in parallel, you can also do the following:

```ts
const compileContractAsyncFns = _.map(contractBaseNames, contractBaseName => {
  return this.compileContractAsync.bind(this, contractBaseName);
});
await Promise.all(compileContractAsyncFns);
```

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+                const opts = {
+                    encoding: 'utf8',
+                };
+                currentArtifactString = await readFileAsync(currentArtifactPath, opts);
+                currentArtifact = JSON.parse(currentArtifactString);
+                oldNetworks = currentArtifact.networks;
+                const oldNetwork: ContractData = oldNetworks[this.networkId];
+                shouldCompile = _.isUndefined(oldNetwork) ||
+                                oldNetwork.keccak256 !== sourceHash ||
+                                oldNetwork.optimizer_enabled !== this.optimizerEnabled;
```
fabioberger: It's considered best practice to keep the contents of a `try/catch` block as small as possible. The reason being that if it throws, you want to know what caused the throw. E.g if the file read succeeds but the `JSON.parse` fails because the contents of the file contained malformed JSON, is simply setting `shouldCompile` to `true` still the correct behavior? Wouldn't it be better to throw a more helpful error message so the developer could fix the issue?

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+            } catch (err) {
```
fabioberger: Let's at least log this error to the console just so that debugging issues here will be easier.

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+                shouldCompile = true;
+            }
+
+            if (shouldCompile) {
```
fabioberger: To avoid excessive conditional nesting, let's do:

```ts
if (!shouldCompile) {
  return; // no-op
}

...
```

This way the rest of this methods logic doesn't live within this `if` block.

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+            let currentArtifact: ContractArtifact;
+            let oldNetworks: ContractNetworks;
+            let shouldCompile: boolean;
+            try {
+                const opts = {
+                    encoding: 'utf8',
+                };
+                currentArtifactString = await readFileAsync(currentArtifactPath, opts);
+                currentArtifact = JSON.parse(currentArtifactString);
+                oldNetworks = currentArtifact.networks;
```
fabioberger: We need to check if `oldNetworks` is `undefined`. Since we cannot guarantee that the contents in the file are properly formatted, it would make this code less brittle to add this check.

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+            }
+        });
+    }
+    /**
+     * Recursively retrieves Solidity source code from directory.
+     * @param  dirPath Directory to search.
+     * @return Mapping of contract name to contract source.
+     */
+    private async getContractSources(dirPath: string): Promise<ContractSources> {
+        let sources: ContractSources = {};
```
fabioberger: It's considered best practice to declare variables as close as possible to where it is used. Let's move this down to right above the `for` loop.

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+        return sources;
+    }
+    /**
+     * Searches Solidity source code for compiler version.
+     * @param  source Source code of contract.
+     * @return Solc compiler version.
+     */
+    private parseSolidityVersion(source: string): string {
+        try {
+            const solcVersion = source.match(/(?:solidity\s\^?)([0-9]{1,2}[.][0-9]{1,2}[.][0-9]{1,2})/)[1];
```
fabioberger: According to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) `String.prototype.match` doesn't throw if no matches are found, but rather returns `null`. We should be checking for `null` instead of having this in a `try/catch`

scripts/Compiler.ts
```diff
@@ -0,0 +1,237 @@
...
+    /**
+     * Normalizes the path found in the error message.
+     * Example: converts 'base/Token.sol:6:46: Warning: Unused local variable' to 'Token.sol:6:46: Warning: Unused local variable'
+     * This is used to prevent logging the same error multiple times.
+     * @param  errMsg An error message from the compiled output.
+     * @return The error message with directories truncated from the contract path.
+     */
+    private getNormalizedErrMsg(errMsg: string): string {
+        try {
+            const errPath = errMsg.match(/(.*\.sol)/)[0];
```
fabioberger: Handle case where `null` returned here.

scripts/Compiler.ts
abandeali1: That's probably cleaner, but currently this throws if trying to access an index of `null`.

scripts/index.ts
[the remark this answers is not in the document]
abandeali1: This will probably become the entry point for deploying as well. I'm thinking `cli.js`.

scripts/Compiler.ts
abandeali1: I guess it depends how we want to handle malformed artifacts. I think it probably makes sense to overwrite them rather than throw an error (ditto for your above comment).

PR #170, opened 2017-11-13
Add custom deployer and migrations

abandeali1:
- Adds a `Deployer` class, which deploys a contract using the ABI in the custom artifacts
- Adds a `migrations` directory, where a custom migration script can be added for use with the CLI
- Adds commands `compile`, `migrate`, and `deploy` to the CLI


deploy/cli.ts
```diff
@@ -0,0 +1,155 @@
...
+import {CompilerOptions, DeployerOptions} from './src/utils/types';
+import {getNetworkIdIfExistsAsync} from './src/utils/network';
+import {commands} from './src/commands';
+
+const DEFAULT_OPTIMIZER_ENABLED = false;
+const DEFAULT_CONTRACTS_DIR = path.resolve('contracts');
+const DEFAULT_ARTIFACTS_DIR = `${path.resolve('build')}/artifacts/`;
+const DEFAULT_NETWORK_ID = 50;
+const DEFAULT_JSONRPC_PORT = 8545;
+const DEFAULT_GAS_PRICE = '20000000000';
```
LogvinovLeon: `(2 * (10 ** 9).toString()`

deploy/cli.ts
```diff
@@ -0,0 +1,155 @@
...
+        type: 'string',
+        default: DEFAULT_CONTRACTS_DIR,
+        description: 'path of contracts directory to compile',
+    })
+    .option('network-id', {
+        type: 'number',
+        default: DEFAULT_NETWORK_ID,
+        description: 'mainnet=1, kovan=42, testrpc=50',
+    })
+    .option('optimize', {
```
LogvinovLeon: What about optimizer runs?

deploy/cli.ts
```diff
@@ -0,0 +1,155 @@
...
+    })
+    .command('compile',
+    'compile contracts',
+    voidCommandBuilder,
+    onCompileCommand)
+    .command('migrate',
+    'compile an deploy contracts using migration scripts',
+    voidCommandBuilder,
+    onMigrateCommand)
+    .command('deploy',
```
LogvinovLeon: Why is this usefull?

deploy/src/compiler.ts
```diff
@@ -15,48 +21,40 @@ import {
...
 const SOLIDITY_FILE_EXTENSION = '.sol';
 
 export class Compiler {
-    private contractsDir: string;
-    private networkId: number;
-    private optimizerEnabled: number;
-    private artifactsDir: string;
-    private contractSources: ContractSources;
-    private solcErrors: Set<string>;
+    private _contractsDir: string;
```
LogvinovLeon: We have this convention only for public interface so that people not using TS but using JS can get the knowledge about stuff being private. Do we plan to have JS users here?

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+import promisify = require('es6-promisify');
+import {
+    ContractArtifact,
+    DeployerOptions,
+} from './utils/types';
+
+const {consoleLog, stringifyWithFormatting} = utils;
+
+export class Deployer {
+    private _artifactsDir: string;
```
LogvinovLeon: Same question. We probably don't need underscores here.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+        const promiWeb3ContractInstance = new Contract(web3ContractInstance, this._defaults);
+        return promiWeb3ContractInstance;
+    }
+    /**
+     * A promisified version of `contract.new`.
+     * @param contract Web3 contract to deploy.
+     * @param args Constructor arguments to use in deployment.
+     * @param txData Tx options used for deployment.
+     */
+    // tslint:disable-next-line:max-line-length
```
LogvinovLeon: Why not break this line?

deploy/src/utils/types.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export interface ContractSources {
+    [key: string]: string;
+}
+
+export interface ImportContents {
+    contents: string;
+}
+
+export enum ZeroExError {
```
LogvinovLeon: Don't we have that exported from 0x.js?

deploy/cli.ts
```diff
@@ -0,0 +1,155 @@
...
+    'deploy a single contract with provided arguments',
+    deployCommandBuilder,
+    onDeployCommand)
```
fabioberger: Ditto above.

deploy/migrations/migrate.ts
```diff
@@ -0,0 +1,13 @@
+import {Deployer} from './../src/deployer';
+
+/**
+ * Custom migrations should be defined in this function. This will be called with the CLI 'migrate' command.
+ * @param deployer Deployer instance.
+ */
+export async function runMigrationsAsync(deployer: Deployer): Promise<void> {
```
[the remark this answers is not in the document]
fabioberger: Yeah, let's make this an object with this single method.

deploy/src/commands.ts
```diff
@@ -0,0 +1,19 @@
...
+import {runMigrationsAsync} from './../migrations/migrate';
+
+export const commands = {
+    async compileAsync(opts: CompilerOptions): Promise<void> {
+        const compiler = new Compiler(opts);
+        await compiler.compileAllAsync();
+    },
+    async migrateAsync(opts: DeployerOptions): Promise<void> {
+        const deployer = new Deployer(opts);
+        await runMigrationsAsync(deployer);
```
fabioberger: This can then be:

```
await migrator.runMigrationAsync(deployer);
```

deploy/cli.ts
```diff
@@ -0,0 +1,155 @@
+import * as yargs from 'yargs';
+import * as path from 'path';
+import {CompilerOptions, DeployerOptions} from './src/utils/types';
```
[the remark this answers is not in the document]
LogvinovLeon: Why? They fit on one line.

deploy/src/utils/fs_wrapper.ts
```diff
@@ -0,0 +1,10 @@
+import * as fs from 'fs';
+import promisify = require('es6-promisify');
+
+export const readdirAsync = promisify(fs.readdir);
+export const readFileAsync = promisify(fs.readFile);
+export const writeFileAsync = promisify(fs.writeFile);
+export const mkdirAsync = promisify(fs.mkdir);
+export const doesPathExistSync = fs.existsSync;
+export const removeDirAsync = promisify(fs.rmdir);
+export const removeFileAsync = promisify(fs.unlink);
```
fabioberger: Let's stick with the convention of having modules either export a class or object. This simplifies the mental overhead of dealing with a new module. Here, this would be re-written as:

```
export const fsWrapper = {
  readdirAsync: promisify(fs.readdir),
  ...
};
```

deploy/src/compiler.ts
```diff
@@ -15,48 +21,40 @@ import {
...
-const readdirAsync = promisify(fs.readdir);
-const readFileAsync = promisify(fs.readFile);
-const writeFileAsync = promisify(fs.writeFile);
-const doesPathExist = promisify(fs.access);
-const mkdirAsync = promisify(fs.mkdir);
+} from './utils/types';
 
-const JSON_REPLACER: null = null;
-const NUMBER_OF_JSON_SPACES = 4;
+const {consoleLog, stringifyWithFormatting} = utils;
```
fabioberger: I'm not a fan of this convention. It makes it harder to grep for all the places where a module is used by grepping for `utils.` or `.consoleLog`. Thoughts @LogvinovLeon @BMillman19

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+import {utils} from './utils/utils';
+import {encoder} from './utils/encoder';
+import {writeFileAsync} from './utils/fs_wrapper';
+import promisify = require('es6-promisify');
+import {
+    ContractArtifact,
+    DeployerOptions,
+} from './utils/types';
+
+const {consoleLog, stringifyWithFormatting} = utils;
```
fabioberger: Don't like this convention. Makes these methods look like globally available methods.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+    constructor(opts: DeployerOptions) {
+        this._artifactsDir = opts.artifactsDir;
+        this._jsonrpcPort = opts.jsonrpcPort;
+        this._networkId = opts.networkId;
+        const jsonrpcUrl = `http://localhost:${this._jsonrpcPort}`;
+        const web3Provider = new Web3.providers.HttpProvider(jsonrpcUrl);
+        this._web3 = new Web3(web3Provider);
+        this._defaults = opts.defaults;
+    }
+
```
fabioberger: Nit: Remove this extra new line.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+     * @param args Array of contract constructor arguments.
+     * @return Deployed contract instance.
+     */
+    public async deployAsync(contractName: string, args: any[]): Promise<Web3.ContractInstance> {
+        const artifactPath = `${this._artifactsDir}/${contractName}.json`;
+        let contractArtifact: ContractArtifact;
+        try {
+            contractArtifact = require(artifactPath);
+        } catch (err) {
+            throw new Error('Contract artifact not found');
```
fabioberger: Let's show them the contractName in the error message:

```
`Artifact not found for contract: ${contractName}`
```

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+        const contractData = contractArtifact.networks[this._networkId];
+        if (_.isUndefined(contractData)) {
+            throw new Error('No contract data found on this network');
```
fabioberger: Ditto above comment. It'll be helpful while debugging to know **which** contract **and** network we didn't find network data for.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+        const data = contractData.unlinked_binary;
+        let from: string;
+        if (_.isUndefined(this._defaults.from)) {
+            const accounts: string[] = await promisify(this._web3.eth.getAccounts)();
```
fabioberger: Why aren't we using `web3Wrapper` here? We should only promisify web3 methods in one place in the whole codebase.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+            from = accounts[0];
+        } else {
+            from = this._defaults.from;
+        }
+        const gasEstimate: number = await promisify(this._web3.eth.estimateGas)({data});
```
fabioberger: Ditto above comment.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+        return promiWeb3ContractInstance;
+    }
+    /**
+     * A promisified version of `contract.new`.
+     * @param contract Web3 contract to deploy.
+     * @param args Constructor arguments to use in deployment.
+     * @param txData Tx options used for deployment.
+     */
+    // tslint:disable-next-line:max-line-length
+    private promisifiedDeploy(contract: Web3.Contract<Web3.ContractInstance>, args: any[], txData: Partial<Web3.TxData>): Promise<any> {
```
fabioberger: Re: above comments:

```
private promisifiedDeploy(
  contract: Web3.Contract<Web3.ContractInstance>, args: any[], txData: Partial<Web3.TxData>,
): Promise<any> {
```

deploy/src/deployer.ts
```diff
@@ -0,0 +1,113 @@
...
+        const deployPromise = new Promise((resolve, reject) => {
+            // tslint:disable-next-line:max-line-length
+            // contract is inferred as 'any' because TypeScript does is not able to read 'new' from the Contract interface
```
fabioberger: Remove tslint:disable and put comment on two lines.

deploy/src/utils/network.ts
```diff
@@ -0,0 +1,14 @@
+import * as Web3 from 'web3';
+import promisify = require('es6-promisify');
+
+export async function getNetworkIdIfExistsAsync(port: number): Promise<number> {
+    const url = `http://localhost:${port}`;
```
fabioberger: Let's move `http://localhost` into a `utils/constants.ts` since it's used in two places.

deploy/src/utils/network.ts
```diff
@@ -0,0 +1,14 @@
...
+    const web3Provider = new Web3.providers.HttpProvider(url);
+    const web3 = new Web3(web3Provider);
+    try {
+        const networkIdIfExists = await promisify(web3.version.getNetwork)();
```
fabioberger: Again, we could use `web3Wrapper` here. Think we shouldn't be using `web3` directly anywhere anymore.

deploy/src/utils/types.ts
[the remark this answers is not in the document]
fabioberger: Once we move 0x.js types into a separate sub-package.

deploy/cli.ts
abandeali1: The function I'm using in solc-js treats this as a flag. It's not clear to me if numbers higher than 1 actually optimize more, but this might be the case.

deploy/cli.ts
abandeali1: If you need to redeploy only a single contract, you can easily do so through the CLI.

deploy/migrations/migrate.ts
```diff
@@ -0,0 +1,15 @@
...
+    /**
+     * Custom migrations should be defined in this function. This will be called with the CLI 'migrate' command.
+     * @param deployer Deployer instance.
+     */
+    async runMigrationsAsync(deployer: Deployer): Promise<void> {
+        // Placeholder for full migrations
+        const tokenTransferProxy = await deployer.deployAsync('TokenTransferProxy', []);
+        const zrxToken = await deployer.deployAsync('ZRXToken', []);
+        await deployer.deployAsync('Exchange', [zrxToken.address, tokenTransferProxy.address]);
+        await deployer.deployAsync('EtherToken', []);
```
fabioberger: @abandeali1 can we perhaps re-implement our existing migrations here? @LogvinovLeon and I both use the migrations when testing 0x.js and if this PR gets merged, we won't be able to migrate the existing contracts anymore.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,115 @@
...
+import {utils} from './utils/utils';
+import {encoder} from './utils/encoder';
+import {fsWrapper} from './utils/fs_wrapper';
+import {
+    ContractArtifact,
+    DeployerOptions,
+} from './utils/types';
+
+// Gas added to gas estimate to make sure there is sufficient gas for deployment.
+const extraGas = 500000;
```
fabioberger: Nit: `EXTRA_GAS`?

deploy/src/deployer.ts
```diff
@@ -0,0 +1,115 @@
...
+        const gas = gasEstimate + extraGas;
+        const txData = {
+            gasPrice: this.defaults.gasPrice,
+            from,
+            data,
+            gas,
+        };
+        const abi = contractData.abi;
+        const contract: Web3.Contract<Web3.ContractInstance> = this.web3Wrapper.getContractFromAbi(abi);
+        const web3ContractInstance = await this.promisifiedDeploy(contract, args, txData);
```
fabioberger: `promisifiedDeploy` should have the `async` suffix. Why not call this: `deployAsync`?

deploy/src/deployer.ts
```diff
@@ -0,0 +1,115 @@
...
+        const newArtifact = {
+            ...contractArtifact,
+            networks: {
+                ...contractArtifact.networks,
+                [this.networkId]: newContractData,
+            },
+        };
+        const artifactString = utils.stringifyWithFormatting(newArtifact);
+        await fsWrapper.writeFileAsync(artifactPath, artifactString);
+        const promiWeb3ContractInstance = new Contract(web3ContractInstance, this.defaults);
```
fabioberger: What does `promi` mean? This can simply be called `contractInstance`. It's not a promise, it's simply a contract instance.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,115 @@
...
+        return promiWeb3ContractInstance;
+    }
+    /**
+     * A promisified version of `contract.new`.
+     * @param contract Web3 contract to deploy.
+     * @param args Constructor arguments to use in deployment.
+     * @param txData Tx options used for deployment.
+     */
+    private promisifiedDeploy(contract: Web3.Contract<Web3.ContractInstance>, args: any[],
```
fabioberger: Should have `async` suffix.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,115 @@
...
+             * is not able to read 'new' from the Contract interface
+             */
+            (contract as any).new(...args, txData, async (err: Error, res: any): Promise<any> => {
+                if (err) {
+                    reject(err);
+                } else if (_.isUndefined(res.address) && !_.isUndefined(res.transactionHash)) {
+                    utils.consoleLog(`transactionHash: ${res.transactionHash}`);
+                } else {
+                    resolve(res);
+                }
```
fabioberger: This body isn't actually `async`. You can drop the `async` prefix to this callback.

deploy/src/deployer.ts
```diff
@@ -0,0 +1,115 @@
...
+        const deployPromise = new Promise((resolve, reject) => {
+            /**
+             * Contract is inferred as 'any' because TypeScript
+             * is not able to read 'new' from the Contract interface
+             */
+            (contract as any).new(...args, txData, async (err: Error, res: any): Promise<any> => {
+                if (err) {
+                    reject(err);
+                } else if (_.isUndefined(res.address) && !_.isUndefined(res.transactionHash)) {
+                    utils.consoleLog(`transactionHash: ${res.transactionHash}`);
```
fabioberger: Hm... I don't think there should be any path that leads to this promise **never** returning. We should probably call `reject` here too.

deploy/src/utils/network.ts
```diff
@@ -0,0 +1,18 @@
...
+    async getNetworkIdIfExistsAsync(port: number): Promise<number> {
+        const url = `http://localhost:${port}`;
+        const web3Provider = new Web3.providers.HttpProvider(url);
+        const defaults = {};
+        const web3Wrapper = new Web3Wrapper(web3Provider, defaults);
+        try {
+            const networkIdIfExists = await web3Wrapper.getNetworkIdIfExistsAsync();
+            return networkIdIfExists;
+        } catch (err) {
+            throw err;
```
fabioberger: If you don't add this `try/catch`, the error thrown will cause the program to exit. No need to re-throw.

deploy/migrations/migrate.ts
abandeali1: The old migrations will still work for now. Maybe this should be in a separate PR? (`yarn migrateCustom` will run the new migrations).

deploy/src/deployer.ts
abandeali1: This exists because the callback for `contract.new` fires twice. This function should never not return, even with the existing code.

deploy/src/deployer.ts
abandeali1: We already have a `deployAsync`. The naming is a bit tricky...

